Issue #56, opened 2025-06-21
[FR] Ability to see task previews in reading mode


amithkk:
Currently, the inline previews for tasks are only for the "live preview" mode and are turned off in reading mode, showing only the wikilinks. Eg.

![Image](https://github.com/user-attachments/assets/e2f30f7b-40f7-418e-8efd-39c8a4c82dfd)

If the inline previews worked for reading mode as well, the behaviour would be more consistent. Willing to contribute a PR!

ky1ejs:
Is this a dupe of #55?

amithkk:
@ky1ejs - I don't believe so. #55 seems to be a bug caused because the overlays were not recreated in "Live Preview" mode when the layout change event occured. 

This is more so around adding the overlays/inline previews in Reading Mode (where the file is not opened for editing) as well.

callumalpass:
Thank you for raising this @amithkk , and for your excellent PR :)

PR #57, opened 2025-06-21
Feat/task previews in reading mode

amithkk:
Addresses #56


callumalpass:
Thanks a lot for this @amithkk. The reading view is something I had completely overlooked. I appreciate the contirbution! :)

Issue #72, opened 2025-06-24
[FR] More compact search/filter bar


amithkk:
I use the tasks view docked on the sidebar (as it is the most compact). However the current card layout with sort/group/filter always open takes up a vertical screen space, and looks a bit out of place with the rest of the obsidian theme

<img width="430" alt="Image" src="https://github.com/user-attachments/assets/1a064f34-2ee5-4f6e-b203-f6334d48b7a6" />

I propose a more compact search/filter bar that looks like what's below:

### Default State - Search only 
<img width="484" alt="Image" src="https://github.com/user-attachments/assets/3e79e137-74a6-4553-baa6-a3bf8c7b5426" />

### With Options Open

<img width="496" alt="Image" src="https://github.com/user-attachments/assets/db4d5fd7-1186-416e-aadd-e3a5db5e037a" />

### With Options Open and Filters expanded

<img width="589" alt="Image" src="https://github.com/user-attachments/assets/e3ab8af0-3425-4652-86a6-1de0b43ec241" />

Willing to contribute a PR :) 

Since this is a common component, the other views like agenda will also be able to leverage the same.

callumalpass:
Hi @amithkk . Thank you for the mock-ups--they look really good! I think this is indeed necessary. I'm working on a refactor of the filter bar at the moment; I'd like to make it both more compact and more powerful. I'm hoping to have a release in the next few days.

callumalpass:
Thank you very much for the PR for this @amithkk, it looks really great!

luciolebrillante:
Hello,

A beautiful improvement! 

Moreover, I think it could be better to have only one mode. When I use filter, I will always have to filter by context because most of the time, I couldn't imagine seeing all my tasks at the same time.

It is painful for me to always have to double click to display those options.

The best filter bar I think will be displayable after one click and you could customise which filter option you want to have in settings in order to display only those you need the most.

bepolymathe:
Thanks @amithkk for this very useful improvement. 
I share @luciolebrillante's point of view because I'm in the same situation. Contexts are very important for filtering everyday tasks that are otherwise too numerous. But I believe that filter features are also set to evolve.

callumalpass:
[...]

PR #73, opened 2025-06-24
Add compact filter bar

amithkk:
Proposed implementation for #72


callumalpass:
Thanks so much for this amithkk. It looks fantastic!

Issue #46, opened 2025-06-16
Feature Request: Status Groups for Kanban


redbeard83:
Hey there!

I've been using Tasknotes with the Kanban feature and it's super helpful. I have a suggestion that I think could make it even better.

Currently, we can create custom statuses for Kanban columns. What would be awesome is to have **"Status Groups"**. This would allow different types of tasks to have their own set of statuses.

For example:

* **Content Creation:** Idea, Writing, Video, Ready, Published
* **Work Tasks:** Backlog, In-Progress, Done
* **Specific Project:** (and so on)

Right now, if I create a context for specific statuses, those statuses only appear when there's a task in them. This means I can't drag a task to a new status if that status is currently empty. Having "Status Groups" would fix this by always displaying all statuses within a chosen group, even if they're empty.

Also, while we can select individual statuses to display, it would be much easier to just select a "Status Group" and have all its associated statuses show up automatically.

One last thing: sometimes when I unselect a status, it still shows up. I have to switch the status group to "Priority" and then back to "Status" for it to disappear. This could be a minor bug or just a usability quirk.

Thanks for considering this! I believe "Status Groups" would make the Kanban feature much more flexible and powerful for managing different workflows.

callumalpass:
Hi @redbeard83 . Thank you for this suggestion! It is an interesting idea. I have a couple of questions:

1. Do you think it is only statuses that need groups? Or do you think there should be something like priority groups as well? 
2. Are you implementing something like this already in your workflow? Am I right in thinking that for each context, you only use certain statuses, and then you're filtering the Kanban board for those different contexts?
3. Would the statuses available for each context or "status group" be stored in the plugin settings. Or would you like to define them in a note somewhere?

redbeard83:
Hi @callumalpass 

Thanks for your questions! Here are my thoughts:

1. Priority groups would indeed be interesting too. I use org-mode for task management, and a great benefit there is being able to create custom statuses and priorities per file. Since TaskNotes uses a file for each task, perhaps grouping files into a "project" and having each project inherit custom statuses/priorities could be a nice solution.

2. Yes, that's right! I use Obsidian to manage content and was trying to implement the same workflow I have in org-mode. 

3. I think it would be easier if these were stored in the plugin settings. Then, you'd simply choose which group/project's statuses to filter or add to a note. For example, when you create a new task, you'd select a group/project for it, and it would inherit the custom statuses of that group/project.

callumalpass:
[...]

PR #236, opened 2025-07-20
feat: Introduce GetGroupedTasksOptions for enhanced task grouping

srcnDev:
- Add GetGroupedTasksOptions interface to support target date and pre-population of groups
- Update FilterService to utilize new options in getGroupedTasks method
- Modify KanbanView to pass options for improved task grouping and display


srcnDev:
On the Kanban board, groups don't appear for statuses that have no tasks, so I made this solution

callumalpass:
Thank you @srcnDev . This is great!

mdbraber:
Could this feature (let groups appear for tasks that have none set) also be implemented for (at least) priorities?

A bit more difficult but this would also be extremely useful for similar situations:
- contexts (we have to define default contexts somewhere)
- dates (today, this week, next week etc) - dragging would set it to the first date of a dynamically defined date range 

This might also fix #46 (if I interpret the request right)

Issue #243, opened 2025-07-20
[FR]: Add classes for tags in the Advanced Calendar


jradam:
UPDATE: Added a pull request for this here: https://github.com/callumalpass/tasknotes/pull/244

### Feature Category

Views (Calendar, Kanban, Agenda, Task List, Notes)

### Problem Statement

Loving the plugin so far! Thank you. 

I would love if there was a way to better differentiate my tasks by tag in the Advanced Calendar.

### Proposed Solution

Could we, in the Advanced Calendar, append a class per-tag to each event? 

For example, if a user had the tags "#meeting" and "#work" on a task, TaskNotes could add classes like `fc-tag-meeting` and `fc-tag-work` to the `<a class='fc-event fc-task-event' />` element.

Then users could write some simple CSS to make all their 'meetings' green and all their 'work' events blue, or whatever.

### Use Case Examples

This would help me see what kinds of tasks I had in a day at a glance. 

### Priority/Impact

None

callumalpass:
Thanks very much for the PR for this, @jradam ! 

I just want to note that the current colours in the Advanced calendar are set by the user-defined priority colours, configurable in the "Priorities" section of the settings. But if users want to control the colours through tags, then this is a great addition!

PR #244, opened 2025-07-20
Iterate through tags and add them as classes to task events

jradam:
Implementation for: https://github.com/callumalpass/tasknotes/issues/243


callumalpass:
Thanks very much for this @jradam !

PR #493, opened 2025-08-26
Feat/Customise Task Card Display Fields (+ filter modal header stays in sync with active view)

renatomen:
## User need
Users can choose which properties appear on each task card so they can see the most important information at a glance without opening the task or navigating to another view.

## Summary
- New: Configurable Task Card Layout (rows 2–4) using a compact token syntax
  - Token format: {property|flags}
  - Implemented flags: 
     - n => Shows the field label. E.g.:
        - {assignee} => Chuck Norris
        - {assignee|n} => Assignee: Chuck Norris
        - {assignee|n(Assigned to)} => Assigned to: Chuck Norris
  - Supports mixing tokens and literal text on the same row. Empty spaces between fields and literals are preserved.
  - 👤{assignee}    🗓️{review_date}
  - Help content is under a collapsible “Layout syntax help” section in the Filter modal
- Improvement: Filter modal header reliably displays the active saved view name
  - Immediately after saving a view
  - When switching/clearing views
  - When loading a tab with a matching view

Demo GIF 
Demo

## Scope of changes
- UI/UX
  - Filter modal: new “Layout syntax help” (collapsed by default, richer guidance and examples)
  - Task card: additional rows after the title (rows 2–4) with small, readable styling aligned with existing metadata
  - Consistent rendering across Task List, Agenda, Kanban, and Project Subtasks widget
- Behavior
  - Rendering uses a lightweight parser for text-based tokens; literal text is preserved
  - Projects support wikilinks (including [[path|alias]]) and render as clickable internal links
  - Filter modal header sync fix: updates the header span without re-rendering the whole modal
  - No breaking changes; default layout shows previous behavior unless a layout is configured.

## Technical rationale
- Approach
  - Simpler, text-based layout syntax avoids complex UI builders, is quick to type, and is easy to review/commit with saved views
  - Parser is intentionally minimal: tokenize “{ … }” segments, split flags, and allow plain text in between
  - Implemented flags limited to n and n(Name) per MVP; possible future extension flags (not implemented) are format flag f(ddd mmm YY).
  - Rendering is capped at three additional rows per card (rows 2–4); title remains row 1
- Cross-view consistency
  - Rendering happens in the shared task card rendering layer and is reused in Task List, Agenda, Kanban, and the Project Subtasks widget
- Filter modal sync
  - On active saved view changes (set/clear or load), we update the header span inside the existing modal (no full re-render), which avoids flicker and maintains state
- Performance impact
  - Very small: parsing and rendering up to 3 light rows per card
  - Token parsing is simple string processing; display value resolution is O(1) per token; array joins only for small lists
  - No additional data fetching beyond existing metadataCache lookups already used elsewhere
- Guidelines adherence (Obsidian plugin best practices)
  - UI separated into modular components; no global state coupling
  - BEM class naming aligned with existing styles (e.g., task-card__field, task-card__custom-row)
  - Optional docs/build pipeline changes only; no intrusive local dev hooks included
  - Backward compatible; new functionality is opt-in

## Known limitations / follow-ups
- Active view restoration still uses “first query match” when loading a tab; storing activeSavedViewId in preferences would eliminate ambiguity when multiple saved views share the same query. This PR doesn’t change that behavior.


callumalpass:
Thank you, @renatomen ! This looks like an incredible contribution!

I want to let you know that this one might take a little longer to merge, and I'd like to explain my thinking on this.

In future releases, I'm hoping to integrate TaskNotes much more tightly with Bases and hand off a lot of the heavy lifting. This might include, for example, much of the work that the filter bar currently does, as well as some other features. I've noted that the Obsidian team has put **grouping** on the roadmap for Bases, as well as Kanban views. Presumably, grouping will make a Kanban view possible in Bases, much like it makes the TaskNotes Kanban view possible.

My expectation is that we will be able to register a "tasknotes" view as part of the Bases plugin API. Presumably, we'll be able to have `.base` files like this:

```yaml
views:
  - type: tasknotes
    name: Task List
    filters:
      and:
        - tags.contains("tasks")
    order:
      - file.name
      - title
      - status
      - ...etc
```

My hope is to fill these views with task cards similar to the ones already used throughout the TaskNotes views. In addition, users will be able to select the properties present on the cards, similar to the current 'Card' view in Bases. You've made the functionality that I had imagined for a future release possible **already** with this PR.

There will certainly be some difficulties when it comes to making Bases compatible with TaskNotes. The biggest complication will involve handling recurring tasks; we'll have to add custom properties that describe their recurrence values. I also foresee that it might be difficult to add context menus to the task cards. Still, I'm hopeful that by leveraging the Bases plugin, we'll be able to simplify the TaskNotes codebase and unlock more powerful features (particularly embeddable views).

My concern--and the reason I'm tempted to wait to merge this PR--is that users will quickly develop complicated saved TaskNotes views that make use of the layout options in this PR. These views will then need to be supported when the Bases plugin API opens up. As part of the future integration with Bases, it would be ideal if we could migrate Saved Views into Bases syntax. Currently, that will involve translating TaskNotes filter conditions into Bases filter conditions. This will be tricky for some filters, but I'm hoping it will be possible. With the functionality you're introducing, we would also have to support the translation of layout rules into Bases properties and property formulas (or, at least, it would be ideal if we could). I'm worried that such a migration might not be possible.

As I see it, we have three options:

1.  Merge now and commit to maintaining compatibility during the eventual Bases migration.
      - 'Maintaining compatibility' could mean either preserving all current TaskNotes views, including the filter bar, or translating the layouts from this PR into Bases property formulas.
2.  Hold off on adding custom properties to the task cards until after the Bases migration.
      - This is the easiest solution, but it's not clear how long we might be waiting for the Bases plugin API.
3.  Modify the proposed implementation in this PR to make future migration to Bases properties easier.
      - For example, we could use checkboxes like those in the Bases "Properties" dropdown or even add minimal support for the property formula syntax.

I'd be interested to hear what you think. I really appreciate the work you've put into this PR--and it has helped me see what might be possible once the Bases API opens up. I'd also be interested to hear whether deeper integration with Bases sounds feasible to you and whether you have any concerns that we might lose something through that integration. I'm still very excited by the functionality you've introduced with this PR and am open to merging it, especially if we can ensure future compatibility with Bases.

renatomen:
Hi @callumalpass, thank you for taking the time to share your perspective.
First and foremost, you don't owe me any justification for your decisions about the plugin. You have the big picture in mind much more than I. As I mentioned before, I'm scratching my own itch with the features I'm implementing, then sharing with you "just in case" it benefits the community and balances value with maintainability.

### I'm glad you mentioned the deeper integration with bases. That was a given in my mind.
- I hope Bases will be flexible enough to support views like the TaskNotes plugin offers. It will be powerful enough to cover most use cases from most users. I'm more hopeful about the filtering and data manipulation it will offer, but also concerned about the level of freedom to develop UIs, such as the one you created in the Tasknotes plugin. 
- Some of the items that I miss a lot in other solutions are 
    - deep nesting... project... task... subtask... subsubtask.... 
    - Ability to have one item associated with multiple parent items (task associated with two projects)
        -  And have it appear under each project when grouping
    - Group and subgroup (e.g. Group by project/subgroup by status) (this one is not implemented in Tasknotes).
 
 So... I worry that bases theming/UI might not support those things very well.
 
 ### About the options you considered
 1) Merging Now and Maintaining Compatibility
     - I'm ready to do that anyway to scratch my own itch. But, my personal needs aside, from a "product" perspective, I wouldn't merge it now unless there's high demand from the community. I'm a super user with very particular needs, trying to bring my workflow to Obsidian. I'm an edge case and, unless the plugin's target users are "edge cases" like me, I wouldn't merge it now. Which makes option 2 - Hold off, likely the right option for now.
 
 2) Option 2 - Hold off
    - Yep. I feel that it will take a while before bases will be ready to "support meaningful Tasknotes features". However, assuming that there's not much pressure to implement this feature, the potential cost of regret is lower if you hold off merging. If you regret and feel that it's taking too long, three months from now, you may be motivated and confident enough that it's not worth waiting for bases. 
    - Conversely. If you merge it now, the bases move fast. There will be a lot of rework involved and user noise.
     
3) Implement now, but make it easier for future migration to bases
    -  I'd defer "preparing" for bases unless that preparation is in terms of fundamental architecture and design patterns. Anything more specific, like syntax or data structure, I'd defer as much as possible until we know what bases will actually offer.
    
4) Possible idea "Tasknotes Labs"
    - 💡One alternative is to merge it and flag it as an experimental feature. It's a common practice to have those features that might not be supported in the future. There could be a Tab in settings called "Labs" with a big warning saying that features enabled there might have bugs, not work correctly and can be discontinued at any time. 
 
Option 4 could be valuable if we have something to learn from it, and users are so desperate that they're interested in the feature even knowing that it might not be there for them tomorrow. 

I don't think that's the case....  (it's the case to me though😁). 

My heart says... merge it now.
My mind says, "Don't merge it. Hold off"! It's not an architecturally mature solution, and it's not convincingly balancing short-term needs and long-term vision well enough.

I didn't know these were my thoughts until you asked. Thanks for asking 😉

renatomen:
Hey @callumalpass ,
After a good night of sleep, I'm changing my mind in relation to my last message. 
It is improbable that the Bases Syntax will radically change. Therefore, I believe it's a good idea to invest into making Tasknotes capable of interpreting Bases syntax. 
Tasknotes approach is deeply reliant on core Obsidian functionality and API, so it makes sense to me to make moves in that direction.

In this case, the idea of an "Experimental" tab in settings could be helpful; we could implement incremental steps towards making Basis the foundation for Tasknotes views.

renatomen:
Hi @callumalpass ,

I'll close this PR. From now on to decluter the PR queue. Any new view configuration / layout feature needs to move "towards" Bases, which is not the case with this one. I'll revisit this later after we agree on some directions and guidelines to our alignment with Bases.

callumalpass:
Thanks for your thoughts here, @renatomen . Sorry for not getting back to you sooner. 

I am interested in implementing an expiremental section to the settings, as you suggested. I think that could be a good way forward. I'd like to look into the funtionality that you implemented with this PR some more. I think the option of adding custom properties to the task cards is in fact very compatible with the Bases syntax, and that we don't need to wait to implement it, especially if we flag it as an experimental feature. I'm not sure that the customization of presentation of their keys will be compatible with the Bases syntax though, and I'm tempted to avoid implementing that part of the PR.

renatomen:
> Thanks for your thoughts here, @renatomen . Sorry for not getting back to you sooner.
> 
> I am interested in implementing an expiremental section to the settings, as you suggested. I think that could be a good way forward. I'd like to look into the funtionality that you implemented with this PR some more. I think the option of adding custom properties to the task cards is in fact very compatible with the Bases syntax, and that we don't need to wait to implement it, especially if we flag it as an experimental feature. I'm not sure that the customization of presentation of their keys will be compatible with the Bases syntax though, and I'm tempted to avoid implementing that part of the PR.

No worries @callumalpass  
I'm excited to share that created a proof of concept. Instead of sharing it with you here, I'll post  the videos in the discussion you opened recently.

Issue #283, opened 2025-07-23
[FR]: Allow multiple instances of specific views


girisumit:
### Feature Category

Views (Calendar, Kanban, Agenda, Task List, Notes)

### Problem Statement

Only one instance of each view can be opened up which might be limiting to some workflows

### Proposed Solution

Have a setting which allows multiple instances of each view.

### Priority/Impact

Medium - This would be a nice improvement

callumalpass:
Thanks @girisumit . I'll look into this.

callumalpass:
Completed in 4.0.1

PR #511, opened 2025-08-27
Feature/kanban instance

noctildon:
related #283 

# Summary
[kanban demo.webm](https://github.com/user-attachments/assets/9320f9fc-ec58-40ad-80dd-afa549931e44)

- Implement embedded kanban view instances
- Create extension .tnkanban

This is experimental. Let me know what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,10 @@
 		"obsidian": "latest",
 		"ts-jest": "^29.1.1",
 		"tslib": "2.4.0",
-		"typescript": "4.7.4"
+		"typescript": "4.7.4",
+		"@types/obsidian-typings": "npm:obsidian-typings@^2.3.4",
+		"@types/react": "^18.2.47",
+		"@types/react-dom": "^18.2.18"
 	},
 	"dependencies": {
 		"@codemirror/view": "^6.37.2",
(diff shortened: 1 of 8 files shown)


renatomen:
OMG @noctildon !!!! Thank you for this fantastic feature.
You read my mind. That was one of my most wanted features.
Great work.
I'm very excited to try it.

@callumalpass and @noctildon, considering the discussion on #497, I think features like this give us a great opportunity to experiment with our ideas. At the same time, it becomes even more important to align our vision, principles, and guidelines with the direction of "aligning" with Bases as you proposed.

I'm making this comment to invite @noctildon to participate in the discussion on #497, where we can have a more specific discussion.

callumalpass:
Thank you very much for this PR, @notctildon!

I'm having some trouble with the behaviour of the Kanban board when it is embedded in notes--it doesn't seem that the cards are updated by dragging them to different boards. 

I'm also a little hesitant to support this new filetype without further consideration. I like the idea of embedding Kanban boards through embedded files, rather than through the use of codeblocks, and I think the file schema is quite nice. I am curious, though, about the Kanban boards that will made available in a future release of Bases, and so I'm not sure I'd like to immediately add support for the schema. I will think on it! Thank again!

noctildon:
@callumalpass Thanks for the feedback. I just fixed it. It should be good now.

My thought is to instantize the views (not only kanban, but also others) and make them embedded in a note. So there might be a number of new extensions (.tnkanban, .tncalendar, etc..) if the development keeps going on.

On the other hand, codeblock doesn't need to register new extensions, which makes the files cleaner.

Certainly future Base with kanban support would make this obsolete, but we don't know how soon it will be implemented.

callumalpass:
I think I will close this PR. Thank you for your work on it @noctildon. But I am keen to have embeddeed views build off of Bases. I would rather not have a custom syntax that we are responsible for maintaining.

PR #512, opened 2025-08-28
Feat/Implement status auto-suggestion with NLP integration

renatomen:
## What

- Added status auto-suggestion triggered by configurable character (statusSuggestionTrigger setting)
- Enhanced NaturalLanguageParser to handle custom status extraction before date parsing
- Fixed suggestion dropdown text truncation with CSS improvements

## Why

- Often users need to add tasks with a status that differ from the default status. E.g when adding a task that's already in progres.
- This feature allows quick status selection during task creation via natural language input
- Prevents conflict when parsing status names containing keywords like 'Now' by parsing status before chrono-node date parser

Demo

## How

- User-defined trigger defined in Settings -> Misc. Defaults to "*".
- Status parsing is disabled if the trigger setting is empty.
- Implemented the NLPSuggest class extending AbstractInputSuggest for status suggestions
- String-based matching;
- Uses case-insensitive indexOf() to handle any characters that Obsidian properties accept
- Processes status configs by length (longest first) to prevent partial matches
- Manual boundary checking instead of unreliable regex word boundaries


renatomen:
Hi @callumalpass , just a heads up, I'm using an enhanced version of the project suggester that's making a huge difference to me.
I have a large vault and just having file name, title and alias in the project auto-suggester didn't give me enough context because I might have the projects or areas with the exact same names in different spaces.

For example, I have three organisations, in all three I have a [[Money, Org & ADM]] Parent, with the same title, name, and aliases.
The only thing that distinguishes them is their path (they're stored in different root folders because physical separation is important in this case for backup and privacy reasons). 

To make life easy, I implemented an enhanced auto-suggester with configurable context. Below is a teaser for you.
I'll wait for you to stabilze the main branch. Then I'll rebase my feature branch for a smoother merge experience.

Of course, I don't expect you to accept all my PRs... you're absolutely free to reject them.

renatomen:
@callumalpass 

- Rebased onto main to harmonize with recently merged PRs, especially the support for multi-word project names.
- Improved code and test quality with better separation of business logic into more testable services, separation from UI.

CHORE: Failing tests cleanup (Not related to this particular branch)
- Removed tests that were incorrectly expecting the overlay functionality that was removed when you merged #505 and simplified the code.

callumalpass:
Thank you for this @renatomen ! This is a nice new feature--well-implemented! Merging :)

Issue #449, opened 2025-08-19
[Bug]: On Task View, respect Obsidian width setting


minchinweb:
### What's the bug?

Obsidian has a setting that will set a "max width" for rendered content. Currently (on the Tasks view at least), the plugin uses the whole available width instead.

### What did you expect to happen?

Have the Tasks view respect the maximum width setting used by Obsidian.

### Which part of TaskNotes is affected?

Task List View

### Screenshots or Screen Recordings

<img width="1667" height="347" alt="Image" src="https://github.com/user-attachments/assets/77f87a33-4cc2-479d-90b9-52d9d6b5f002" />

callumalpass:
Thanks for this, @minchinweb . I'll see what can be done!

Issue #105, opened 2025-06-29
[Bug]: Task title overflows margins on mobile


cathywu:
### Obsidian Debug Info

N/A

### Bug Description

1. Create inline task with a long title
2. Go back to the note from which the task is created
3. View the note on Obsidian mobile app
4. Pan right and see that the text goes beyond the margins

### Expected Behavior

The text should not go beyond the margins.

### Actual Behavior

It does go beyond the margins

### Feature Context

Task Management (creating, editing, status changes)

### Screenshots

Example failure:
![Image](https://github.com/user-attachments/assets/e5087436-4d84-40b0-a91a-69b2204142bc)

callumalpass:
Thanks for the report. I could not reproduce this overflow on the latest release.

The inline task card styles now include mobile-specific constraints for inline cards: the inline card is capped to the note width, the row/content/title elements can shrink with `min-width: 0`, and long title text is clipped with ellipsis rather than expanding the note horizontally.

I think this can be closed as fixed/current unless it still reproduces on the latest release with a small note fixture, theme name, and device details.

PR #517, opened 2025-08-29
fix(css): enable wrapping for inline task overlay title via white-space

renatomen:
# Problem
Inline tasks overlay wasn't wrapping in small screens and extended the container beyond screen boundaries, forcing hoirixontal scroll.

image

# Solution
Changed style attribute from nowrap to break-spaces
image


callumalpass:
Thanks @renatomen ! This should fix a long-standing issue (#105) ! Very much appreciated!

callumalpass:
Also fixes #449

Issue #470, opened 2025-08-21
[Bug]: Due date doesn't update for recurring tasks


jhedlund:
### What's the bug?

I am testing out TaskNotes, I created a task that is due tomorrow 8/22, repeats on Thursday (scheduled 8/21).

When I complete the task, the task updates to 8/28 (next Thursday) correctly, but the due date remains 8/22. 

### What did you expect to happen?

I would have expected the due date to move by the same number of days as the schedule date, due date of 8/29 in this case.

### Which part of TaskNotes is affected?

None

callumalpass:
Thank you for reporting this jhedlund. Recurring tasks currently do not pay much attention to due dates. The behaviour that you're suggesting here sounds quite sensible. I'll have a think about implementing it.

jhedlund:
Thanks! I started looking at it a bit as well, I can try to create a PR unless you have a specific way you want to accomplish it.  I see where the update to the scheduled property is for the recurring tasks.  

I tried running unit tests but several timezone tests failed (and a couple others), not sure if the unit tests are supposed to run successfully or not.

PR #522, opened 2025-08-29
Allow selection of date field for recurrence rule to apply, automatically move other date field with same offset

jhedlund:
This is to address #470

Somewhat of a work in progress, but is working with my manual testing.

This adds an option in Custom Recurrence modal to select which date field the rule should apply to (scheduled by default, due date is another option).   I realized that some of my tasks I want to recur based on the due date and not the scheduled date which may be before the due date.

If scheduled date is the recurring date field, then the due date will automatically adjust based on the offset between the original scheduled and due date.

Likewise, if due date is the recurring date field, then the scheduled date will automatically adjust based on the offset between the original scheduled and due date.

Let me know if this is a suitable approach for you or if you would prefer a different direction.


callumalpass:
Thank you very much for submitting a PR, @jhedlund . I'm a little hesitant to merge it for a couple of reasons. 

First, I'd like to avoid adding new elements to the data schema. Adding a new property is a significant change, and I'd like to only add properties where they provide significant functionality. 

You suggest that this PR fixes the offset issue that you opened--that it should update both scheduled and date values. I'm not seeing this behaviour. I think I'd like to get this behaviour fixed first before giving users control over whether the recurrence rules apply to the due or scheduled dates. 

Thank you again though for this PR. The changes are thoughtful, and are very much appreciated!

jhedlund:
Fair enough, this PR does fix the issue (of course with additional feature), the fix is focused in helpers.ts, where it computes the offset between the due date and the scheduled date.

I can reduce this PR to just focus on fixing #470 by moving the due date by the offset from the original scheduled date and the due date. 

As I was coding that fix, I realized that for some tasks (perhaps most?), the due date is what I'm looking to repeat.   For example, I have tasks that I want to repeat on the due date every 3 months on the 1st of the month.   Ideally I'd want to repeat that on the due date but set the scheduled date a week or two in advance to see that it's a scheduled task but due on the 1st.

jhedlund:
Addressed #470 solo in #542, closing this PR for now

PR #542, opened 2025-08-31
Move Due Date when recurring task completed

### LINKED ISSUE #470: [Bug]: Due date doesn't update for recurring tasks

jhedlund:
Moves due date when recurring task completed  by the same offset as the original due date is to the original scheduled date to keep same separation

Fixes #470

### DIFF
--- a/src/modals/TaskEditModal.ts
+++ b/src/modals/TaskEditModal.ts
@@ -548,9 +548,12 @@ export class TaskEditModal extends TaskModal {
             // If task has recurrence, update scheduled date to next uncompleted occurrence
             if (this.task.recurrence) {
                 const tempTask: TaskInfo = { ...this.task, ...changes };
-                const nextScheduledDate = updateToNextScheduledOccurrence(tempTask);
-                if (nextScheduledDate) {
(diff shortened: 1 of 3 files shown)


callumalpass:
Thank you very much for this contribution, @jhedlund . I've made these changes configurable because it might cause unexpected behaviour for some users. I've very much appreciate your work here--it resolves a long-standing issue (mentioned by @jpmoo [here](https://github.com/callumalpass/tasknotes/discussions/237#discussion-8608406))

jhedlund:
Excellent, thank you for merging!

PR #555, opened 2025-09-01
Enhanced “+” Project Autosuggest: configurable cards, selective search, and clearer matches

renatomen:
## Why this is needed (succinct justification)
Users often have many similarly named project notes. A single-line filename list forces guesswork, slows creation, and leads to wrong selections. Showing a compact, configurable card (e.g., Title, Aliases, Path) makes the right note obvious at a glance. Allowing selective search across chosen fields (via a simple flag) balances precision and performance in large vaults.

## Example (what this unlocks)
- Typing “+ acme” can now match a project even if “acme” appears in a chosen field like customer or file path, while the card clearly shows:
  - Title: Acme Corp Project
  - Aliases: acme, client-acme
  - Path: Clients/Acme/Project.md

Users immediately see they’re selecting the correct note and can tune which fields are searchable.
Demo

## What’s included
- Configurable multi-row project cards in the “+” autosuggest
- Simple field flags:
  - n or n(Label): show field name before value
  - s: include that field in “+” search (in addition to filename/title/aliases defaults)
- Optional fuzzy search for broader, multiword matching
- Unit tests covering parsing, metadata resolution, rendering, highlighting, and “+” search with flagged fields
- Help docs updated with examples and two short demo GIFs

## How to use
- Settings → Misc → Project Autosuggest:
  - Define up to 3 rows (mix literals and fields like title, aliases, file.path) - Only displayed as metadata, not considered as searchable fields.
  - Add |s on fields you want included in “+” search. This considers the flagged fields as searcheable, therefore if the field contains the search term the result is considered a positive match.

## Performance and safety
- Defaults remain fast and safe
- Search scope expands only when users add |s on specific fields
- Capped suggestions and early exit paths keep large vaults responsive

## Tests: Use cases covered by this PR’s test suite
(Many tests below were initially broken by the package and security change in the latest release and required refactoring)

- Display-fields parser
  - Parses {property|flags} tokens, including n and n(Label)
  - Supports |s searchable flag
  - Handles literals and escaping (pipes, parens)
  - Serializes tokens back to strings

- “+” project search behavior
  - Defaults searchable fields work: filename (basename), title (via mapping), aliases
  - |s adds extra searchable fields on top of defaults (e.g., file.path)
  - |s enables matching custom frontmatter fields (e.g., customer)
  - Respects per-row configuration (up to 3 rows)
  - Optional fuzzy mode supported

- Project cards rendering
  - Renders multi-row cards with configured rows (mix of literals and fields)
  - Shows labels when n/n(Label) is present
  - Correctly joins array values (e.g., aliases)
  - Includes filename row plus configured rows

- Selective highlighting
  - Highlights only searchable fields (filename/title/aliases always; |s-marked fields additionally)
  - Works with mixed rows (searchable + non-searchable tokens)
  - Highlights matches inside array elements
  - Works with and without “+” trigger context

- Metadata resolution
  - Resolves file.* (basename, name, path, parent)
  - Resolves title, aliases, frontmatter keys
  - Renders wikilinks/markdown links as plain text
  - Handles missing/empty fields gracefully

- Trigger detection
  - Detects @, #, + suggestion triggers at the correct cursor position
  - Ignores non-trigger contexts

- Status suggestions (regression coverage)
  - “*” status autosuggest remains functional and returns expected options


renatomen:
Hi @callumalpass 
This one was ready a while ago and I was waiting for you to merge the status autosuggester to save you the trouble of merging the two. 
Unfortunately (or fortunately) with so many updates in the last release a lot of changes were required, especially in the unit tests.
No complains. That's a good thing. Only extra work compared the usual.

There are many other failing tests from other feature after the latest release. I'll refactor them later and submit manageable dedicated PRs for that.

The Bases integration will require a similar effort because it's been implemented in the previous release. 
Not a problem. It will just take a bit more time.

I've been stuck in a Ski Camp in the past few days because the mountain is closed due to bad weather, so I've been having fun with TaskNotes ⛷️😁.

callumalpass:
Thank you @renatomen! This looks very impressive. I'll give it a look over and hopefully merge it soon. It looks like it will be very helpful for solving the issues that people have been having with filtering their projects. This is an issue that I haven't really had any ideas for fixing, some I'm grateful for your contribution here. 

I hope the weather improves for you--even if it means that we see fewer of your fantastic contributions here!!

renatomen:
Thank you @callumalpass.
Yoohoo! I'm so glad you merged this one and added the ability to filter. I was going to work on that next.
There's one thing I miss not only here but in other places like the filter bar, which is the ability to leverage Obsidian's hierarchical tags system, the ability to exclude tags (and the same with folders).

I raised a FR explaining in more detail. In the same FR I volunteer to submit PRs implementing those requests.
https://github.com/callumalpass/tasknotes/issues/584

Issue #494, opened 2025-08-26
[Bug]: Can not move tasks beween kanban lists when grouped by custom field


anareaty:
### What's the bug?

I hoped to be able to use custom fields for kanban lists but it doesn't work. When I group kanban by the custom field I get the lists, but when I try to drag a task between them I get an error "Failed to move task". I tried it with text and list fields with the same result.

### Which part of TaskNotes is affected?

None

callumalpass:
Thanks @anareaty . I'll look into a fix for this. Please note that this may not be fixed right away, and that I'm hoping to deprecate the Kanban view in a future release, in favour of a native Bases Kanban view (mentioned [here](https://obsidian.md/roadmap)). See also #497

callumalpass:
Given some of the limitations of Bases, it's looking like the Kanban view will not be deprecated. I will try to get a fix for this in 3.25

callumalpass:
This should be fixed in 3.24.0

PR #565, opened 2025-09-02
Feat/clean bases integration on top of 3.22.1 + All Bases Test Passing + No Console Errors + Advanced Data Logs

renatomen:
Hi @callumalpass 

As I mentioned in my last comment:

- Rebased onto the latest main
- Has no console errors
- Has advanced data login (a data dump that gives you the ability to see all data provided by bases - you need to enable another toggle in settings in addition to the POC logs)
- Dependency update: Obsidian-Typings updated to the latest version (release a couple of days ago).

I encourage you to try the advanced log below
image

image

⚠️ There are multiple test unrelated to Bases that are still failing after lots of changes were introduced by the latest plugin releases. As I mentioned before, I'll address them separately on a dedicated PR to main. 

It might be easier to just replace this entire branch with my branch.

Later... because tomorrow the weather looks great and I'll finally go to the mountain ⛷️

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "tasknotes",
-	"version": "3.21.0",
+	"version": "3.22.1",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "tasknotes",
-			"version": "3.21.0",
+			"version": "3.22.1",
 			"license": "MIT",
 			"dependencies": {
 				"@codemirror/view": "^6.37.2",
@@ -36,6 +36,7 @@
 				"jest": "^30.1.1",
 				"jest-environment-jsdom": "^30.1.1",
 				"obsidian": "latest",
+				"obsidian-typings": "^4.49.0",
 				"ts-jest": "^29.1.1",
 				"tslib": "^2.8.1",
 				"typescript": "^5.9.2"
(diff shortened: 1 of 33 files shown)


callumalpass:
Thanks for this @renatomen , it is incredibly helpful! I'll work on integrating it with the rest of the codebase over the next few days. 

Just so you're aware, I think I'd like to aim to avoid adding custom syntax to the Bases views, at least until after the Bases API is released. I am hopeful that we can push minimal support for Bases out shortly, and that users can have confidence that the Bases that they create will be compatible with future Bases releases (even if the Bases views risk malfunctioning in the near future because of changes to the internal Bases API). I love the `groupBy` options you've demonstrated, but because they rely on custom syntax, I think we should hold out on supporting this until we can see the syntax that the Obsidian team decides on. 

Thanks again for all your amazing work on this, and enjoy the snow!

renatomen:
Hi @callumalpass ,
The groupBy is not custom. You can use that syntax in a fresh obsidian vault with zero plugins.
The syntax is official and I got it in discussions from obsidian discord group.

renatomen:
@callumalpass while the syntax is solid, they removed that from the official docs because the UI required more work and there are other priorities now. So it makes sense to leave it out if the UI settings to avoid an avalanche of requests and bugs.

You can't set groups via UI, but you can edit the bases file and it works (with terrible UI).

Since tasknotes doesn't rely on bases view for grouping, that's reasonably safe to use that syntax. 

The worse that can happen is that they will change the groupBy keyword again ,(it already changed). But at this point, anything can change anyway....

Cheers

callumalpass:
Oh, that's very interesting! Okay, that gives me a lot to think about!

I was hoping that the Bases API would help us deprecate the Kanban view. But playing around with the internal API, I suspect that the task cards by themselves won't be "registerable" with Bases. I think we'll have to be implementing entire views, rather than just implementing row/note-level cards. Which means that whatever Kanban view the Obsidian team puts together, we won't be able to add our own TaskNote cards to it. If that 'groupBy' key is fairly stable, then we probably do have enough to safely implement the Kanban view as well! 

In which case, I'll try to get working on some of the Kanban issues (like #494)

renatomen:
@callumalpass 
We can ask the Obsidian team about their views on registering "row level" elements in the Dev channel.

renatomen:
@callumalpass 
Closing this base integration pull request, as I submitted this one before you publicly released the integration. This PR is obsolete now. I'm happy that the integration went from idea to reality much sooner than we thought 🎇🙌

Issue #354, opened 2025-08-07
[Bug]: Cursor show above Subtasks Widget when placement is "Top of note"


mdbraber:
### What's the bug?

When selecting "Top of note" for "Project subtasks widget position" it shows the cursor blinking *above* the subtasks widget. As soon as I start typing the text goes *below* the subtasks widget as expected.

### What did you expect to happen?

The cursor should be blinking *under* the subtasks widget when selectiong "Top of note" as the subtasks widget position

### Which part of TaskNotes is affected?

None

callumalpass:
Thanks @mdbraber . This is actually one of the reasons why I put off making it possible to move the subtasks widget to the top of the note. It produces some very strange behaviour, at times. I'll have to spend a bit more time trying to figure out how to implement it better.

mdbraber:
No problem - maybe as a pointer - the Virtual Content Plugin also does something similar. Maybe there's something to take from their logic e.g. https://github.com/Signynt/virtual-content/blob/68b56bde5f7958ef0c3bdc9dfc1e12aa6e2a4ce2/main.ts#L672 (or @signynt might be able to help?)

callumalpass:
This should be resolved in v4. Thank you for the pointer to the Virtual Content Plugin, @mdbraber--it's a useful reference for Tasknotes users. In v4, the relationships widget and the taskcard widget both use a similar approach to the Virtual Content Plugin--DOM injection--rather than a Codemirror widget.

PR #625, opened 2025-09-10
Remove cursor above subtasks when subtasks are set to show at top of project note

jhedlund:
I like to have my subtasks at the top of my project notes, but I noticed that there is a cursor shown above the subtasks but jumps to below the subtasks when you start typing.  This leaves a line gap and looks confusing.

This css fixes the issue by setting the CodeMirror widgetBuffer img tag to display: none; if the project-note-subtasks are at the top (or specifically, not at the bottom).

I tested this with both top and bottom display of subtasks and it worked well in my tests.  Oddly enough, if I set it to display: none while at the bottom, the opposite problem occurs (a cursor appears below the subtasks), that is why I look for not last-child.

### DIFF
--- a/styles/project-note-subtasks.css
+++ b/styles/project-note-subtasks.css
@@ -36,6 +36,11 @@
     outline: none;
 }
 
+/* Hide CodeMirror widgetBuffer img when it's adjacent to our widget and our widget is on top of note (not last-child) */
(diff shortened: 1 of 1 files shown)


jhedlund:
Just saw there is a issue for this, closes #354

callumalpass (approved):
Thanks very much for this @jhedlund. This has bothered me for a long time!

PR #640, opened 2025-09-12
fix: should not contains brackets in project

### LINKED ISSUE #638: [Bug]: {{project}} template conversion should not contains `[[` and `]]`
### What's the bug?

Actually, when the default folder template contains a project, as it is a link, the brackets are not removed.

### What did you expect to happen?

The brackets should be removed

### Which part of TaskNotes is affected?

Task Management

### Screenshots or Screen Recordings

_No response_

### Technical Details (Optional)

_No response_

Mara-Li:
Close #638


callumalpass (approved):
Thank you @Mara-Li!

PR #658, opened 2025-09-14
feat: Add "Hide subtasks" setting to UI Elements options.

needo37:
## Summary

Adds a "Hide subtasks" setting that filters subtasks from all views while keeping them accessible via parent task expansion.

## Key Features

- **New setting**: Toggle in Appearance & UI → UI Elements
- **Universal filtering**: Works across Task List, Kanban, and Agenda views
- **Smart detection**: Recognizes both `[[Task]]` wikilinks and plain text project references
- **Non-destructive**: Subtasks remain expandable under parent tasks

## Technical Details

- New `TaskRelationshipUtils` class for relationship detection
- Integrated into `FilterService` for seamless filtering
- Full TypeScript support with proper error handling
- Follows existing project coding standards

## Usage

Enable "Hide subtasks" in settings to hide subtasks from main views while preserving their functionality as expandable subtasks.

### DIFF
--- a/src/services/FilterService.ts
+++ b/src/services/FilterService.ts
@@ -8,6 +8,8 @@ import { FilterUtils, FilterValidationError, FilterEvaluationError, TaskProperty
 import { isDueByRRule, filterEmptyProjects, getEffectiveTaskStatus } from '../utils/helpers';
 import { format } from 'date-fns';
 import { splitListPreservingLinksAndQuotes } from '../utils/stringSplit';
+import { TaskRelationshipUtils } from '../utils/TaskRelationshipUtils';
+
 import {
     getTodayString,
     isBeforeDateSafe,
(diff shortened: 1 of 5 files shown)


callumalpass:
Thank you very much for this PR, @needo37 . It is very much appreciated. I think that I will not merge this, however, because users can already hide subtasks from views by using the "Projects is empty" filter condition. This seems to me to be a more flexible way of achieving the functionality that this PR implements, and avoids the maintainence overheads of a new setting. 

Thanks again!

needo37:
Sure, I appreciate that logic. However @callumalpass I feel that just changing that filter is not enough. And maybe I am doing it wrong and am open to hearing a better workflow.

I have a PARA method where I have a top level "Projects" directory with all of my projects. Not all of these projects have subtasks. I keep them tagged as project not task.

In order to get TaskNotes to recognize it I do tag a project that is going to have multiple tasks underneath it as project and task. When I do this though everything shows up twice in the Kanban board for example. If I set the "Project is empty" filter I then see the projects that have subtasks. But I don't see any of my other tasks that aren't or don't have a subtask. I cannot figure out how to see all my tasks and not have duplicates in one pane of glass. That's where this PR came from.

I am happy to adjust my workflow. I am just at a loss.

Issue #672, opened 2025-09-16
[Bug]: Subtask Widget doesn't show up


Volker-brdb:
### What's the bug?

After installing the lastest version (3.23.2) of TaskNotes-Plugin
the project subtask widget doesn't show up anymore.
Even after a restart of obsidian.

Option is activated:

<img width="777" height="116" alt="Image" src="https://github.com/user-attachments/assets/cf8265a2-f9a1-4b23-8098-bd4e0cc46f32" />

### Which part of TaskNotes is affected?

None

callumalpass:
Thank you very much for reporting this @Volker-brdb --very much appreciated! This is now fixed in 3.23.3

Issue #674, opened 2025-09-16
[Bug]: Expandable Sub-Tasks are gone from Kanban view


JD2026:
### What's the bug?

The Expandable sub-task option is on but the sub-task do not appear in the Kandan view after recent version update 

### Which part of TaskNotes is affected?

None

anareaty:
The same problem, I also don't see them in tasklist and agenda views anymore.

callumalpass:
Thank you for reporting this @anareaty and @JD2026 . I'll push a fix shortly.

callumalpass:
This is fixed in 3.23.3

PR #675, opened 2025-09-16
Fix project/subtask functionality to respect user field mapping

callumalpass:
## Problem
ProjectSubtasksService and MinimalNativeCache were hardcoded to look for `projects` in frontmatter, ignoring user's custom field mapping configuration. This broke subtask/project functionality when users customized their project field name.

## Solution
- Use `fieldMapper.toUserField('projects')` instead of hardcoded `'projects'`
- Apply fix to both ProjectSubtasksService and MinimalNativeCache

## Files Changed
- `src/services/ProjectSubtasksService.ts`
- `src/utils/MinimalNativeCache.ts`

Fixes reported issue where subtask/project functionality wasn't working for users with custom project field mappings.


callumalpass:
To fix #672 and #674

Issue #447, opened 2025-08-19
[FR]: Allow "required task" (rather than only subtasks)


minchinweb:
### Feature Category

None

### Problem Statement

Sometimes, a task only becomes active when a "pre-task" is completed. There, allow defining prerequisite tasks for each tasks.

### Proposed Solution

Likely a frontmatter key, defining prerequisite tasks.

Ideally, tasks without their prerequisites complete should be able to be filtered out and hidden.

### Use Case Examples

This will also be useful for implementing Gantt charts (see #438)

### Priority/Impact

Medium - This would be a nice improvement

callumalpass:
Thanks @minchinweb . This is a good suggestion. This will be a major feature (it will involve adding a new field), but I agree it would be helpful to have.

callumalpass:
Related to #701 . This will be implemented in 3.24

renatomen:
@callumalpass , 

This is related to [[FR]: Mindmap view for child/parent relations](https://github.com/callumalpass/tasknotes/issues/730).

Recently in the TaskNotes community, there has been a growing interest in "Task Dependencies". From memory, I remember
- "Task Sequencing"
- "blockedBy" task
- "Required task" (this FR)
- "Gantt Dependencies" (Start to Start, Start to Finish, Finish to Start, Finish to Finish)

Just like calendars adhere to RFC standards, maybe it's a good idea to consider similar standards to support dependencies.
In [my Gantt Plugin](https://github.com/callumalpass/tasknotes/issues/704) I didn't decide yet how dependencies will be stored in frontmatter. When I get to that poing, I'll check for standards that could be useful.

I have three main requrements:
- It must be easy for a human to understand by checkng the frontmatter.
   - Which wouldn't be the case if I stored JSON or other complex syntax in frontmatter
- It must be easily consumable by other plugins.
- It would ideally comply or be very aligned with widely used open standards for dependency management.
   - E.g.: RFC 5545/9253, and be "round-trippable" to iCalendar standards with One-to-one mappings for VTODO/RRULE and RELATED-TO with RELTYPE and GAP.

Of course, I will let users map their properties to the ones expected by the plugin instead of enforcing specific property names.

There's potential overlap between blockedBy, senquencing that I haven't thought much about. But my approach would be to stick to the open standards and try to resolve all "task dependency / senquencing" use cases with open standards... then see if there's any edge case that requires tweaking those standards.

Below is an "idea" (not a plan) of what my Gantt plugin might expect

[...]

PR #734, opened 2025-09-27
feat: implement blocking/blocked task dependencies

callumalpass:
## Summary
Implements blocking/blocked task dependency relationships with proper persistent architecture.

- Tasks can be marked as blocking other tasks or blocked by other tasks
- Dependency relationships persist in frontmatter as structural links
- Dynamic blocked state computed based on blocking task completion status
- Clean UI integration in task modals and task cards

## Key Features
- **Persistent relationships**: `blockedBy` links stored permanently in frontmatter
- **Dynamic state**: `isBlocked` computed in real-time from blocking task statuses  
- **UI integration**: Task cards show blocked badges, modals allow dependency editing
- **Performance optimized**: Single file write per status change, efficient UI updates

## Technical Implementation
- Added `blockedBy`/`blocking` fields to TaskInfo type and field mapping
- Enhanced task modals with dependency editing interface
- Updated cache system to track dependency relationships and compute blocked state
- Added context menu actions for quick dependency management
- Implemented proper UI update triggers for dependent tasks

## Test Plan
- [x] Create tasks with blocking/blocked relationships
- [x] Verify dependency links persist correctly in frontmatter
- [x] Confirm blocked state updates when blocking tasks are completed
- [x] Test UI shows correct blocked badges and dependency lists
- [x] Validate no data loss when tasks are completed/reopened


renatomen:
Hi @callumalpass ,

I didn't have the chance to check how dependencies were implemented here. I just left a comment about what I'm considering for task dependency management in the gantt plugin. 
here's the link to the comment https://github.com/callumalpass/tasknotes/issues/447#issuecomment-3342206645

Cheers👋

callumalpass:
Thanks for pointing me to the 9253 standard, @renatomen . I was hoping to push a release in the next few hours of the "blockedBy" feature, but I'll have a think about whether it would be prudent to follow that standard and maybe hold off. he 9253 looks like it could be quite powerful!

In the current implementation (in this branch), the "blockedBy" frontmatter key is just a link to another task that is blocking the current task. There will be new filter conditions to filter out blocked/blocking tasks. A task is considered blocked if the task that it points to in its "blockedBy" property is incomplete. 

I'll have a think about the different relationships there might be, though, and that 9253 standard describes.

callumalpass:
@renatomen, I'm really liking the four different temporal relationships that the specification describes (FINISHTOSTART, FINISHTOFINISH, STARTTOFINISH, STARTTOSTART). 

I'd be keen to adopt your suggestion by making the "blockedBy" field look something like:

```
blockedBy: # This will obviously be configurable in the field mapper
  - uid: "[[path/to/Task A]]"
    reltype: FINISHTOSTART
    gap: P1D
  - uid: "[[path/to/Task B]]"
    reltype: STARTTOSTART
---
```

The only complication will be the requirement to define which statuses count, for the user, as "started". Users can currently define some statuses as "completion" statuses--and this does a lot of work throughout the plugin. In order to properly support "STARTTOFINISH" and "STARTTOSTART" we'd have to add an option for users to define certain statuses as "started". 

Are you thinking that you will require users to define a "started" status for the GANTT plugin?

callumalpass:
An alternative to using a "started" status would be to make the "scheduled" date count as the "start" date...

PR #744, opened 2025-09-27
Add task card widget in task notes

### LINKED ISSUE #576: [FR]:  Show task context menu when the current note is a task
### Feature Category

None

### Problem Statement

Many times I double-click to open a task to get to it's actual note so I can add extra notes or information to the task.

### Proposed Solution

I see there is a command palette command for "Quick actions for current task", but it would be much easier and less clicks if I could just right-click in the note and access the task context menu.  

Or perhaps there another way I can do this faster to have actions available for the current task? 

Some examples would be to start the timer easily, complete a task, change the status, etc - pretty much the normal right-click context menu and also the options in the quick actions for current task, just making those more readily accessible.

### Use Case Examples

_No response_

### Priority/Impact

None

### Additional Context

_No response_

callumalpass:
## Summary

Implements a configurable task card widget that displays at the top of task notes, providing immediate access to task actions and context menus without requiring the command palette.

## Changes

- Add TaskCardNoteDecorations extension using CodeMirror widget system
- Task card displays at top of note after frontmatter when present
- Uses existing createTaskCard function for consistency with other views
- Add showTaskCardInNote setting (enabled by default)
- Ensure task card appears above project subtasks widget when both enabled
- Minimal styling with border and small padding
- Responds to task updates, deletions, and settings changes

## Implementation Details

- CodeMirror WidgetType for rendering task card
- Automatic detection of task files using existing cache manager
- Live preview mode only (not shown in source mode)
- Event-driven updates for real-time synchronization
- Proper cleanup and memory management

## Addresses

Closes #576

Eliminates the need to use the command palette for "Quick actions for current task" by providing direct access to task context menus and interactive elements within the task note itself.

## Test Plan

- [ ] Verify task card appears in task notes when enabled
- [ ] Confirm task card does not appear in non-task notes
- [ ] Test setting toggle functionality
- [ ] Verify positioning above project subtasks widget
- [ ] Confirm real-time updates when task properties change
- [ ] Test in both light and dark themes
- [ ] Verify proper cleanup when switching between notes

### DIFF
--- a/build-css.mjs
+++ b/build-css.mjs
@@ -21,6 +21,7 @@ const CSS_FILES = [
     'styles/unscheduled-tasks-selector-modal.css', // UnscheduledTasksSelectorModal component with proper BEM scoping
     'styles/task-action-palette-modal.css', // TaskActionPaletteModal component with proper BEM scoping
     'styles/project-note-subtasks.css',  // ProjectNoteSubtasks component with proper BEM scoping
+    'styles/task-card-note-widget.css',  // TaskCardNoteWidget component with proper BEM scoping
     
     // BEM View Files
     'styles/task-list-view.css',     // TaskListView component with proper BEM scoping
--- a/src/editor/TaskCardNoteDecorations.ts
+++ b/src/editor/TaskCardNoteDecorations.ts
@@ -1,0 +1,353 @@
+import { Decoration, DecorationSet, EditorView, PluginSpec, PluginValue, ViewPlugin, ViewUpdate, WidgetType } from '@codemirror/view';
+import { EVENT_DATA_CHANGED, EVENT_TASK_DELETED, EVENT_TASK_UPDATED, EVENT_DATE_CHANGED, TaskInfo } from '../types';
+import { EventRef, TFile, editorInfoField, editorLivePreviewField } from 'obsidian';
+import { Extension, RangeSetBuilder, StateEffect } from '@codemirror/state';
+
+import TaskNotesPlugin from '../main';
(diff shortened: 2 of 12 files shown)


aubreyz:
What about Reading Mode - which as for the subtasks widget is a lot more relevant to most people than live preview. I suspect most people toggle between reading mode and source mode, and only use live preview for specific editing tasks (such as tables) and for previews.

callumalpass:
I'll be looking in to adding support for the reading view, @aubreyz . It requires a very different implementation though, and is not a straightforward fix. My understanding is that most users *only* use the live preview (I've had to ask users in the past to use source mode to solve some problems they've encountered, and it has been a shock for them to discover that source mode exists).

aubreyz:
> I'll be looking in to adding support for the reading view, @aubreyz . It requires a very different implementation though, and is not a straightforward fix. My understanding is that most users _only_ use the live preview (I've had to ask users in the past to use source mode to solve some problems they've encountered, and it has been a shock for them to discover that source mode exists).

There was a recent Reddit thread on this very topic with widely differing viewpoints. 
https://www.reddit.com/r/ObsidianMD/comments/1l1o3lz/what_use_does_reading_view_has_over_live_preview/

Lots of people don't use Live Preview at all because (per that thread) a) fear of inadvertent changes b) More importantly because a lot of things don't render properly in live preview

That said, if it is a whole different coding, then you are working extremely hard and have more important things to do (including in the rest of your life...)

callumalpass:
That's an interesting thread--thank you!

Issue #907, opened 2025-10-12
[Bug]: "Add to Project" in the Task Edit modal still using simple tag matching instead of hierarchical matching. Results are inconsistet with the Task Creation modal.


renatomen:
### Bug Description

When I click on "Add to Project" button in the Task Edit modal, the project suggestions do not return the same results as the Task Creation modal.
While the Task Creation modal correctly returns results filtered by the plugin's "project auto-suggest" settings respecting hierarchical tags and exclusion settings, the "Task Edit" modal still filters based on simple tags.

**Expected Behavour**
- Project suggestion in Task Creation and Task Editing are consistent (return the same results)
- Project suggestion filter in the Task Editing modal returns results consistent with the Project Auto-Suggestion settings.

callumalpass:
Thanks very much for this PR @renatomen.

Issue #905, opened 2025-10-12
[Bug]: Project auto-suggest filters are unexpectedly applied to custom field file selection


renatomen:
### Bug Description

When editing a cutom field in the Task Edit modal, the file suggestions are filtered by the plugin's "project auto-suggest" settings. 

In other words, for all custom fields, the file auto-suggestion shows the same results as the project auto-suggestion.

**Expected Behavour**
- File suggestions in custom fields are not filtered by the project autosuggestion settings.

But originally reported by @bepolimathe in the discussion issue https://github.com/callumalpass/tasknotes/discussions/143

callumalpass:
Resolved in 3.25.3. Thanks @renatomen for reporting this. Closing this issue.

PR #908, opened 2025-10-12
fix: Prevent project filters from affecting custom field suggestions / Inconistent Project Suggestions (Bugs #905, #907)

renatomen:
## Problem

**Bug**: Project autosuggestion filter settings (tags, folders, properties) were incorrectly applied to custom field file/link suggestions when users typed `[[` in custom fields.
Bug: https://github.com/callumalpass/tasknotes/issues/905

**Bug**: When I click on "Add to Project" button in the Task Edit modal, the project suggestions do not return the same results as the Task Creation modal.

While the Task Creation modal correctly returns results filtered by the plugin's "project auto-suggest" settings respecting hierarchical tags and exclusion settings, the "Task Edit" modal still filters based on simple tags.
Bug: https://github.com/callumalpass/tasknotes/issues/907

**Impact**: Users could not select files that didn't match their project filter criteria when using wikilink suggestions in custom fields, even though custom fields should show all vault files.

## Root Cause

1. **FileSuggestHelper** always applied project filters to ALL callers, regardless of context
2. **ProjectSelectModal** (used by Task Edit modal) used simple tag matching instead of hierarchical matching with exclusion support

This caused:
- Custom field wikilink suggestions to only show filtered project files instead of all vault files
- Task Edit modal to show fewer projects than Task Creation modal (missing files that matched via hierarchical tags)

## Solution

Made filtering **opt-in and generic**:
- Helper is now filter-agnostic
- Callers explicitly decide which filters to apply
- Both modals use consistent hierarchical tag matching
- Future-proof for custom field-specific filters

### `src/suggest/FileSuggestHelper.ts`
**What changed**: Added generic `FileFilterConfig` interface and made filtering optional

**Why**: 
- Previously always read from `plugin.settings.projectAutosuggest` and applied filters to all callers
- Now accepts optional `filterConfig` parameter - `undefined` means no filtering
- Generic interface not tied to "project" concept, reusable for future custom field filters

**Key change**:
```typescript
// Before: Always applied project filters
const requiredTags = plugin.settings?.projectAutosuggest?.requiredTags ?? [];

// After: Only apply if filterConfig provided
const requiredTags = filterConfig?.requiredTags ?? [];
```

### `src/modals/TaskCreationModal.ts`
**What changed**: Explicitly passes project filter config to FileSuggestHelper

**Why**: 
- Project field (`+` trigger) should filter by project settings
- Makes intent explicit: "I want project filtering"
- No longer relies on helper's default behavior

**Key change**:
```typescript
// Explicitly pass project filter configuration
const list = await FileSuggestHelper.suggest(
    this.plugin,
    queryAfterTrigger,
    20,
    this.plugin.settings.projectAutosuggest  // Explicit filter config
);
```

### `src/modals/TaskModal.ts`
**What changed**: Calls FileSuggestHelper without filterConfig for custom fields

**Why**: 
- Custom fields (`[[` trigger) should show ALL vault files
- No filterConfig parameter = no filtering
- Clear intent: "Show me everything"

**Key change**:
```typescript
// No filterConfig = shows all files
const list = await FileSuggestHelper.suggest(this.plugin, partial);
```

### `src/modals/ProjectSelectModal.ts`
**What changed**: Updated tag filtering to use `FilterUtils.matchesTagConditions()`

**Why**: 
- Previously used simple `.some()` matching (exact matches only)
- Now uses same hierarchical tag matching as FileSuggestHelper
- Supports exclusion patterns (tags with `-` prefix)
- Consistent behavior across Task Creation and Task Edit modals

**Key change**:
```typescript
// Before: Simple exact matching
const hasRequiredTag = requiredTags.some((reqTag) => allTags.includes(reqTag));

// After: Hierarchical matching with exclusion support
if (!FilterUtils.matchesTagConditions(allTags, requiredTags)) {
    return false;
}
```

### `tests/unit/suggest/FileSuggestHelper.test.ts`
**What changed**: Added 12 unit tests covering filtering behavior

**Why**: 
- Verify no filterConfig returns all files
- Verify filterConfig applies filters correctly
- Test tag, folder, and property filtering independently and combined
- Ensure query matching works regardless of filter settings

**Test coverage**:
- Filter configuration (with/without filterConfig)
- Tag filtering (filtered vs unfiltered)
- Folder filtering (filtered vs unfiltered)
- Property filtering (filtered vs unfiltered)
- Multiple filters combined
- Query matching

### DIFF
--- a/src/modals/ProjectSelectModal.ts
+++ b/src/modals/ProjectSelectModal.ts
@@ -10,6 +10,7 @@ import type TaskNotesPlugin from "../main";
 import { ProjectMetadataResolver } from "../utils/projectMetadataResolver";
 import { parseDisplayFieldsRow } from "../utils/projectAutosuggestDisplayFieldsParser";
 import { getProjectPropertyFilter, matchesProjectProperty } from "../utils/projectFilterUtils";
+import { FilterUtils } from "../utils/FilterUtils";
 
 /**
  * Modal for selecting project notes using fuzzy search
@@ -56,7 +57,7 @@ export class ProjectSelectModal extends FuzzySuggestModal<TAbstractFile> {
 
 			const cache = this.app.metadataCache.getFileCache(file);
 
-			// Apply tag filtering - use native Obsidian API
+			// Apply tag filtering - use FilterUtils for consistent hierarchical tag matching
 			if (requiredTags.length > 0) {
 				// Get tags from both native tag detection and frontmatter
 				const nativeTags = cache?.tags?.map((t) => t.tag.replace("#", "")) || [];
@@ -68,9 +69,8 @@ export class ProjectSelectModal extends FuzzySuggestModal<TAbstractFile> {
 						: [frontmatterTags].filter(Boolean)),
 				];
 
-				// Check if file has ANY of the required tags
-				const hasRequiredTag = requiredTags.some((reqTag) => allTags.includes(reqTag));
-				if (!hasRequiredTag) {
+				// Use FilterUtils.matchesTagConditions for hierarchical matching and exclusion support
+				if (!FilterUtils.matchesTagConditions(allTags, requiredTags)) {
 					return false; // Skip this file
 				}
 			}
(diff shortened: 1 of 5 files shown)


callumalpass:
Thank you @renatomen ! Much appreciated!

PR #955, opened 2025-10-17
Fix/953 subtasks incorrectly identified as task notes

### LINKED ISSUE #953: [Bug]: All TaskNotes Views include subtasks that don't match the "Task Identification" settings
### Bug Description

Given TaskNotes Plugin setting maps "Project" to a property named "Parent"
AND a Tasknotes "Task Identificatioin" -> Task Tag is set to "t/ef"

WHEN I create a note titled "**Random_Note**"
AND the **Random_Note** DOES NOT contain any tags
AND the **Random_Note**'s "Parent" property value includes a link to a an existing "TaskNotes Task" titled "**Existing_Task**"
THEN 👇

#### **Unexpected Behaviour**
⛔The Random_Note is displayed as a subtask of "**Existing_Task**"

#### **Expected Behaviour**
👍The Random_Note IS NEVER displayed as a TaskNote result as a Task, nor Subtask

#### **Reason for the expected behaviour**
Makes Task Identification inconsistet. Violates the "Task tag" setting which determines  "how Tasknotes identifies notes as tasks"

[Demo Video](https://i.ibb.co/WNnd1shj/Obsidian-o3q3p-FPKOd.gif)

#### Impact on workflow
In vaults where projects and tasks can be a "Parent" of tasks and non-task notes, the non-task notes pollute the TaskNotes resutls. This is even worse for large projects with lots of documents.

In my workflow, the "Parent - Child" relationship is determined by the "in" property.
E.g: 
🚧Project A
     |_ ☑️Task A1
     |_ ☑️Task A2
     |_ ☑️Task A3
     |_ 📝Note NA1
     |_ 📝Note NA2
     |_ 📝Note NA3

When managing tasks, I only want to see Task A1, Task A2, Task A3 under Project A.

For users with workflows in where seeing documents as subtasks is desirable, they can always add a tag to the document to make it "visible" in their TaskNotes results.

The current behaviour offers no easy way to exclude no-task notes from subtasks results.

renatomen:
## Description

Fixes #953 - Prevents non-task notes from being incorrectly identified as subtasks and ensures views update instantly when task identification changes.

## Root Causes

1. `MinimalNativeCache.getTaskInfo()` returned `TaskInfo` for any file with frontmatter, without validating task identification criteria
2. `ProjectSubtasksService.buildProjectIndex()` marked tasks as projects based on links from non-task files
3. Cache invalidation events weren't firing when files stopped being tasks (early exit prevented event emission)
4. `ProjectSubtasksService` had no event listeners for metadata changes, relying only on 30-second TTL
5. **Cache events weren't propagating to plugin emitter** - `MinimalNativeCache` triggered `file-updated` events but `EVENT_DATA_CHANGED` was never fired, so views didn't refresh
6. **Task validation missing in extraction** - `extractTaskInfoFromNative()` extracted TaskInfo without validating task status, causing asymmetric behavior (task→note vs note→task transitions)

### Task Identification Validation
- Added `isTaskFile()` validation to `getTaskInfo()` method
- Made `isTaskFile()` public for use by `ProjectSubtasksService`
- Added `isTaskFile()` validation to `buildProjectIndex()` method
- Created comprehensive test suite with 13 test cases

###  Event-Driven Cache Invalidation
- Added `initialize()` method to `ProjectSubtasksService` with event listeners
- Track files that were previously tasks to ensure events fire on task→non-task transitions
- Clean up cached data when files stop being tasks
- Invalidate project index immediately on metadata changes

###  Test Mock Fixes for CI
- Fixed test failures in GitHub Actions by properly mocking TFile and dependencies
- Created mockFile using `Object.create(TFile.prototype)` to pass instanceof checks
- Added missing parameters and methods to mocks

### Cache Event Bridge & Task Validation
- **Cache Event Bridge** (`main.ts`): Added `setupCacheEventBridge()` method to propagate cache events (`file-updated`, `file-deleted`, `file-renamed`) to plugin emitter's `EVENT_DATA_CHANGED`, ensuring views refresh when cache changes
- **Task Validation** (`MinimalNativeCache.ts`): Added `isTaskFile()` validation to `extractTaskInfoFromNative()` to ensure it returns `null` when file stops being a task, preventing early exit in comparison logic
- Uses `requestAnimationFrame` for efficient UI updates
- Fixes asymmetric behavior where views updated instantly when note→task but not when task→note

## Impact

- Non-task notes no longer appear as subtasks
- Chevron only appears when actual task subtasks exist
- Views refresh **immediately** when task identification changes (both tag-based and property-based)
- Works in both directions: note→task AND task→note transitions
- No need to wait 30 seconds or restart Obsidian for changes to take effect
- No breaking changes - only corrects invalid behavior and completes missing event flow

## Testing

- 13 unit tests for task identification validation (tag-based, property-based, edge cases)
- 5 unit tests for cache event bridge functionality
- All MinimalNativeCache tests passing 
- All cache event bridge tests passing
- Verified non-task notes don't appear as subtasks
- Verified instant view updates when task identification changes

## Files Modified

- `src/utils/MinimalNativeCache.ts` - Task validation in getTaskInfo() and extractTaskInfoFromNative()
- `src/services/ProjectSubtasksService.ts` - Event listeners and cache invalidation
- `src/main.ts` - Cache event bridge setup
- `tests/unit/utils/MinimalNativeCache.getTaskInfo.test.ts` - Test suite (13 tests)
- `tests/unit/main.cacheEventBridge.test.ts` - New test file (5 tests)


callumalpass:
Thanks for this fix for #953, @renatomen ! However, I'm worried the event-driven index invalidation will cause performance issues. Every file edit triggers a full subtask index rebuild (scanning all files), which can take 1-2 seconds in large vaults. The subtask index rebuilding can be very slow, which is why it is only allowed every 30 seconds. 

I think I might look into a different fix for #953 in v4. Basically, I think I'll add a view option that allows users to select whether filter conditions flow down into subtasks. This should allow the appropriate filtering of non-task notes from the subtasks.

renatomen:
Hi @callumalpass, that's a good point. I didn't experience performance issues, but other users might.
In relation to the comment below:

> I think I'll add a view option that allows users to select whether filter conditions flow down into subtasks. This should allow the appropriate filtering of non-task notes from the subtasks.

That's a good idea, and it could work depending on how you implement it. 

I'm also concerned with the possibility that it might not solve #953. The solution you're describing originally addresses a different concern. Which is about filter consistency between tasks/projects and subtasks.

My concern is that if I don't choose the option to force filter conditions to flow down to subtasks (essentially keeping the current behaviour) the incorrect task identification would remain.

The filter function requires notes to be identified as "TaskNotes" as the primary data source. Once identified, a note could appear as either a task or a subtask based on workflow preferences (flow down or not)

However, bug #953 is a separate concern, independent of filters. It breaks TaskNotes validation by allowing users to identify TaskNotes through an unofficial method. Currently, there are two valid identification methods:

A) Tags-based
B) Property-based

Bug #953 creates a third unofficial way: assigning a TaskNote project to a non-TaskNote note. This makes TaskNote identification unreliable.

Thoughts?

renatomen:
@callumalpass, here's some food for thought in case you think it is worth resolving the task identification issue in isolation.

Below are two possible options to address inconsistent Task identification while minimizing or eliminating side effects (e.g., excessive index rebuilds). 

### Original PR Problem

1. **Cache Event Bridge** (`main.ts`): Triggers `EVENT_DATA_CHANGED` on every `file-updated` event, causing all views to refresh
2. **ProjectSubtasksService Event Listeners**: Invalidates the project index on every `file-updated` event by setting `indexLastBuilt = 0`

The combination means:
- User edits a file → `file-updated` event fires
- ProjectSubtasksService sets `indexLastBuilt = 0`
- Views refresh via `EVENT_DATA_CHANGED`
- Views call `isTaskUsedAsProjectSync()`
- Index is rebuilt because `indexLastBuilt = 0` (bypassing the 30-second TTL)
- **This happens on EVERY file change**, defeating the purpose of the 30-second TTL


### Alternative Solution 1: **Remove Event-Driven Invalidation Entirely**


- Remove ProjectSubtasksService.initialize() event listeners
- Keep only the 30-second TTL
- Keep task validation in buildProjectIndex()
- **Trade-off**: Views may show stale project/subtask data for up to 30 seconds after changes

### Alternative Solution 2: **Selective Project Index Invalidation**

The cache event bridge is actually already somewhat selective (MinimalNativeCache only fires `file-updated` when changes affect task rendering). The real problem is that ProjectSubtasksService invalidates on ALL task-related changes, even when project relationships haven't changed.

**Solution**: Make ProjectSubtasksService invalidation selective by only invalidating when changes actually affect project relationships.

#### Implementation Approach

**1. Track Project-Relevant State**

Add state tracking to ProjectSubtasksService:

```typescript
export class ProjectSubtasksService {
    private plugin: TaskNotesPlugin;
    
    // Existing fields...
    private projectIndex = new Map<string, boolean>();
    private indexLastBuilt = 0;
    private readonly INDEX_TTL = 30000;
    
    // NEW: Track project-relevant state for change detection
    private lastKnownProjects = new Map<string, any>();
    private lastKnownTaskStatus = new Map<string, boolean>();
    
    // ...
}
```

**2. Implement Selective Invalidation Logic**

```typescript
/**
 * Initialize the service and set up event listeners for selective cache invalidation
 */
initialize(): void {
    // Listen to cache events to selectively invalidate the project index
    this.plugin.cacheManager.on("file-updated", (data: { path: string; file: TFile }) => {
        this.handleFileUpdated(data.path);
    });

    this.plugin.cacheManager.on("file-deleted", (data: { path: string }) => {
        this.handleFileDeleted(data.path);
    });

    this.plugin.cacheManager.on("file-renamed", (data: { oldPath: string; newPath: string }) => {
        this.handleFileRenamed(data.oldPath, data.newPath);
    });
}

/**
 * Handle file updates with selective invalidation
 * Only invalidate if project relationships or task status changed
 */
private handleFileUpdated(path: string): void {
    const file = this.plugin.app.vault.getAbstractFileByPath(path);
    if (!(file instanceof TFile)) return;

    const metadata = this.plugin.app.metadataCache.getFileCache(file);
    const frontmatter = metadata?.frontmatter;
    
    // Get current state
    const projectsFieldName = this.plugin.fieldMapper.toUserField("projects");
    const currentProjects = frontmatter?.[projectsFieldName];
    const currentIsTask = this.plugin.cacheManager.isTaskFile(frontmatter);
    
    // Get last known state
    const lastProjects = this.lastKnownProjects.get(path);
    const lastIsTask = this.lastKnownTaskStatus.get(path);
    
    // Check if project-relevant changes occurred
    const projectsChanged = !this.projectsEqual(currentProjects, lastProjects);
    const taskStatusChanged = currentIsTask !== lastIsTask;
    
    // Update tracked state
    this.lastKnownProjects.set(path, this.deepClone(currentProjects));
    this.lastKnownTaskStatus.set(path, currentIsTask);
    
    // Only invalidate if project relationships changed
    if (projectsChanged || taskStatusChanged) {
        this.invalidateIndex();
    }
}

/**
 * Handle file deletion - always invalidate as links are affected
 */
private handleFileDeleted(path: string): void {
    this.lastKnownProjects.delete(path);
    this.lastKnownTaskStatus.delete(path);
    this.invalidateIndex();
}

/**
 * Handle file rename - always invalidate as paths are affected
 */
private handleFileRenamed(oldPath: string, newPath: string): void {
    this.lastKnownProjects.delete(oldPath);
    this.lastKnownTaskStatus.delete(oldPath);
    this.invalidateIndex();
}

/**
 * Compare projects arrays for equality
 */
private projectsEqual(a: any, b: any): boolean {
    if (a === b) return true;
    if (!a && !b) return true;
    if (!a || !b) return false;
    
    if (!Array.isArray(a) || !Array.isArray(b)) {
        return JSON.stringify(a) === JSON.stringify(b);
    }
    
    if (a.length !== b.length) return false;
    
    const sortedA = [...a].sort();
    const sortedB = [...b].sort();
    
    return sortedA.every((val, idx) => val === sortedB[idx]);
}

/**
 * Deep clone helper for tracking state
 */
private deepClone(obj: any): any {
    if (obj === null || obj === undefined) return obj;
    return JSON.parse(JSON.stringify(obj));
}
```

**3. Keep the Task Validation in buildProjectIndex()**

The existing validation added by the PR should be kept:

```typescript
// Validate that the source file is actually a task (issue #953)
// Only tasks should be able to create project relationships
if (!metadata?.frontmatter) continue;
if (!this.plugin.cacheManager.isTaskFile(metadata.frontmatter)) continue;
```

**4. Keep the Cache Event Bridge**

The cache event bridge in `main.ts` should remain as-is because:
- It's already selective (only fires when task rendering changes)
- It ensures views refresh when task data changes
- The performance issue is in ProjectSubtasksService, not the bridge

### Benefits of This Solution

- **Achieves PR's original goal**: Views update instantly when task identification changes (via cache event bridge)

- **Prevents non-task files as subtasks**: Task validation in `buildProjectIndex()` ensures only actual tasks create project relationships

- **Minimizes performance impact**: Index only rebuilds when project relationships actually change, not on every file edit

- **Respects 30-second TTL**: For files where projects don't change, the TTL prevents unnecessary rebuilds

- **Handles all edge cases**:
- Task → non-task transition: Detected by `taskStatusChanged`
- Non-task → task transition: Detected by `taskStatusChanged`
- Project field changes: Detected by `projectsChanged`
- File deletion/rename: Always invalidates (correct behavior)

### Trade-offs

**Complexity**: Adds state tracking and comparison logic to ProjectSubtasksService

**Memory**: Stores last known projects and task status for each file (minimal overhead)

**Correctness**: Relies on accurate change detection (mitigated by comprehensive comparison logic)

callumalpass:
Thanks very much for your thoughts here, @renatomen. I can see how my suggested changes won't resolve the issue. 

Do you think we could just use the changes you implemented in commit 1067495?

renatomen:
> Do you think we could just use the changes you implemented in commit [1067495](https://github.com/callumalpass/tasknotes/commit/106749592f44b4dba4865264c72159b9564196fe)?

## Short answer: 
- Yes. With Caveats.

## Long answer

About the caveats....

I believe it's a good compromise.
- Minimal change with no impact to performance.

Negatives:
- Doesn't update when a TaskNote becomes a regular note. (I suppose it would require closing and reopening the view).
- If a non-TaskNote is linked to the note it would display a chevron and a "No subtasks found" nested under it. (I'm not sure, but I believe that's what motivated me to implement the second commit. Check the image below).

image

This "Chevron" without actual subtasks may annoy users.
I'd rather live with the side effect of this chevron than having the core bug of non-TaskNotes being displayed in TaskNotes views.

renatomen:
@callumalpass , I believe a minimal solution without including instant cache updates would be:

- Implement commit 1067495 ✅ (safe, no side effects)
- Also implement the task validation from commit 2 (https://github.com/callumalpass/tasknotes/pull/955/commits/51f6c32647f4472711ab5c6a05e63eccb3acb49f)  in ProjectSubtasksService.buildProjectIndex():
- Skip the event-driven invalidation from commits 2 (https://github.com/callumalpass/tasknotes/pull/955/commits/51f6c32647f4472711ab5c6a05e63eccb3acb49f) & 4 (https://github.com/callumalpass/tasknotes/pull/955/commits/4de347700a37f8410969c88e9f90046d0b95cf0b) (the performance concern)

This gives you:

👍 Core bug fix (non-tasks won't be identified as tasks)
👍 Project index correctness (non-tasks won't create project relationships)
👍 Zero performance impact (no event listeners)
❌ Views still update on 30-second TTL (acceptable trade-off)

renatomen:
@callumalpass, I committed a simplification refactor (https://github.com/callumalpass/tasknotes/pull/955/commits/771adee69ff8f9ce0898ce2ba638d5723d9ed1dc)

### Removed 
❌ ProjectSubtasksService.initialize() - event listeners that invalidate index on every file change
❌ setupCacheEventBridge() in main.ts - triggers EVENT_DATA_CHANGED on every file edit
❌ MinimalNativeCache "wasTask" tracking - unnecessary complexity
❌ Cache event bridge test file


### Kept (Core Fixes - Zero Performance Impact):

✅ Task validation in MinimalNativeCache.getTaskInfo()
✅ Task validation in ProjectSubtasksService.buildProjectIndex()
✅ Task validation in MinimalNativeCache.extractTaskInfoFromNative() 
✅ isTaskFile() made public for cross-service validation
✅ test suite (refactored)

### Impact
✅ What Works:

- Non-task notes are no longer incorrectly identified as subtasks (issue #953 fixed)
- Only actual tasks can create project relationships
- No performance impact 

⚠️ Trade-off:

- Views update on 30-second TTL instead of instantly
- The chevron + No subtasks found message only appears until the 30-second TTL expires. If the filter is re-applied (e.g. changing the text in the search bar, the chevron behaves as expected after 30s)

callumalpass:
Thank you reworking this, @renatomen . Very much appreciated!

callumalpass:
I've been working on some changes which conflict with this PR, so I can't directly merge it. But I've incorporated--and am very grateful for--the changes that you've made here!

PR #975, opened 2025-10-23
fix: use theme color for advanced calendar list headers

hugohil:
fullcalendar defaults to `var(--fc-page-bg-color)` which doesn't account for the theme color (i.e dark theme has white table headers)


callumalpass:
Thank you very much for this PR, @hugohil . Apologies for taking so long to merge it. Very much appreciate your work here.

PR #1084, opened 2025-11-07
Fix word wrapping in "nl-preview-item" div

maybetm:
Hi ☺️
Small CSS fix in the creation modal – fixed word-wrap for `nl-preview-item` div.

fix-word-wrap-div


callumalpass:
Thanks very much for this PR, @maybetm . Much appreciated! I'm making some big changes to the NLP field for the next beta, so this might be overridden, but I appreciate your efforts here nonetheless!

PR #1090, opened 2025-11-09
Feat add Portuguese Brazil Translations

andysteel:
I'm a Brazilian developer fascinated with the woking doing until now with this plugin. So i decided to translate to portuguese for other people of my country that don't know about english enough  start to use this awesome Solution.
Thanks for all developers that works hard here.


callumalpass:
Thanks so much for this @andysteel ! really appreciate your work here!

PR #1095, opened 2025-11-10
chore: 🔧 Enable maintainDueDateOffsetInRecurring by default

0xSnib:
Related: [https://github.com/callumalpass/tasknotes/discussions/1081](https://github.com/callumalpass/tasknotes/discussions/1081)

maintainDueDateOffsetInRecurring is false by default, I think having the due dates on a task that is recurring not update is confusing to new users (it bamboozled me)


callumalpass:
Thanks for raising this, @0xSnib. I agree the current default can be confusing for new users.

That said, changing a default affects all existing users who haven't explicitly set this option — their recurring task behavior would silently change on update. I think the right approach is to keep the current default but improve discoverability. I'd be open to changing the default in a major version bump where breaking changes are expected.

Issue #791, opened 2025-09-30
[FR]: UUID v4 as filename format option


bangbangbanggit:
### Feature Description

I'm not sure but I checked the available variables: {title}, {titleLower}, {titleUpper}, {titleSnake}, {titleKebab}, {titleCamel}, {titlePascal), {date), {shortDate), (time), {time12}, {time24}, (timestamp}, {dateTime), (year), {month), {monthName}, {monthNameShort}, {day}, {dayName}, {dayNameShort}, {hour}, {hour12}, {minute), {second}, {milliseconds}, {ms}, {ampm], {week}, {quarter}, {unix}, {unixMs}, {timezone}, {timezone Short}, {utcOffset), {utcOffsetShort}, {utcZ}, {zettel), {nano), {priority}, {priority Short}, (status}, {statusShort}, {dueDate), {scheduledDate}

It seems none of the custom variables will generate a UUID v4. If you decide to have this option, get it like from here: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID or like from this code

```
generateUUID(): string {
    return crypto.randomUUID();
}
```
Have the UUID v4 as an option, not in the "Custom filename template" as a variable but in the drop down of "Filename format"

callumalpass:
Thank you for reporting this. This should be resolved in release 4.8.0, so I am going to close this issue now.

Issue #59, opened 2025-06-22
[FR] Short year and abbr month for custom filename template


Pixeltica:
Hiya, would it be possible to add a short year (25 instead of 2025, like the zettel) format and abbreviated month (Jan, Feb, Mar, etc) format as options to the custom filename options?  i.e something like: 
shortYear: format(date, 'YY')
abbrMonth: format(date, 'MMM')
It's mostly cosmetic but would be greatly appreciated! Thanks! :)

callumalpass:
Closing as implemented. https://tasknotes.dev/settings/task-defaults/

PR #1178, opened 2025-11-20
feat: Filename Options (projectId)

bendavis987:
## Background
I made some custom changes for my workflow and figured I could share if you're interested. The point of this PR is really to gauge your interest, and I can polish them up if so. I also wasn't sure if you had already finished: #791 or #59, so I did those as well while I was in these files.

## Filenames move to General Tab
I didn't know the file name configuration was an option initially. This was part of why I started making changes. I can revert this if you prefer it in appearance, but it just didn't seem like it fit there. It should be in the same place as custom folder structure in my opinion.

## Filename context vs taskData structure
I apologize if this is bad practice. I don't know much about TS and it's conventions, but this seemed like it made more sense to me. Maybe filenamecontext should extend it. I'm not sure. Probably worth a conversation.

## Actually what I wanted done (New Filename Option: projectId)
I really wanted to format the task names similar to JIRA ie: TASK-1234, and if it is related to a project, then it'll get the id from the project. PROJ-1234. This is a bit difficult since there currently isn't a tracker for the file numbering incrementing. I decided to just piggyback off of the duplicate name check for now. For a full feature, this should probably get a menu option to configure starting number, and it can be stored there / in the obsidian workspace somehow.
### Related:
- #791: UUID v4 custom filename variable
- #59: Short year

## Other notes
I had a few questions while debugging. Like I broke the translations somehow, I need to figure that out. And the projects don't seem to just be the project name. I can't cut the strings consistently at the same place.

### DIFF
--- a/src/services/TaskService.ts
+++ b/src/services/TaskService.ts
@@ -206,6 +206,7 @@ export class TaskService {
 					title: title,
 					priority: priority,
 					status: status,
+					projects: projectsArray,
 					due: taskData.due,
 					scheduled: taskData.scheduled,
 				},
--- a/src/settings/tabs/generalTab.ts
+++ b/src/settings/tabs/generalTab.ts
@@ -193,6 +193,12 @@ export function renderGeneralTab(
 					),
 				},
 				{
+					value: "project",
+					label: translate(
+						"settings.general.taskFilenames.filenameFormat.options.project"
+					)
+				},
+				{
 					value: "custom",
 					label: translate(
 						"settings.appearance.taskFilenames.filenameFormat.options.custom"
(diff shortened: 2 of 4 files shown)


callumalpass:
Thanks for sharing this, @bendavis987 — the JIRA-style naming is an interesting workflow idea.

As you noted in the PR description, this is still a work in progress (broken translations, inconsistent project name parsing). There are also some structural changes (moving settings between tabs, changing the FilenameContext/taskData relationship) that I'd want to think through carefully.

I'm going to close this for now since it's not quite ready. If you'd like to revisit it later with a more focused scope (just the projectId filename variable, without the settings reorganization), I'd be happy to take another look.

Sorry for the long wait on feedback!

PR #1197, opened 2025-11-21
fix: Add Portuguese language to list of translations for general tab settings

andysteel:
This configs is necessary for the portuguese transalation to be able for use


callumalpass:
Thank you for this, @andysteel ! Merging

PR #1202, opened 2025-11-21
fix: Update task property on kanban drag & drop

### LINKED ISSUE #1200: [Bug]: Dropping a task into "done" in Kanban fails to add `completedDate` to the task
### Bug Description

I found that after updating to TaskNotes v4, dragging a task in Kanban to the status column marked as Completed only changes the status but doesn't add the `completedDate` property to the task, and the usual popup `Task marked as 'Done'` doesn't appear. Additionally, the task is never auto-archived. This issue does not occur when clicking the checkmark button or right-clicking and selecting the 'Done' status.

TaskNotes version: 4.0.3
Obsidian version: 1.10.3

Image

tvoklov:
Fixes #1200 by checking if kanban column/swimlane is a valid task property and calling `updateProperty` on it.

Also fixes the auto-archiving issue, since `updateProperty` puts tasks into the auto-archive queue.

### DIFF
--- a/src/bases/KanbanView.ts
+++ b/src/bases/KanbanView.ts
@@ -874,14 +874,14 @@ export class KanbanView extends BasesViewBase {
 		const task = await this.plugin.cacheManager.getTaskInfo(taskPath);
 		const taskProperty = this.plugin.fieldMapper.lookupMappingKey(frontmatterKey);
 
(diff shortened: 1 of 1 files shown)


callumalpass:
Thanks so much for this fix @tvoklov . It looks like a few people have been running into this issue. Much appreciated!

PR #1206, opened 2025-11-22
fix: Handle status trigger prefix in NLP parsing

renatomen:
## Summary

Fixes NLP status parsing regressions where:
1. Manually typing status trigger + label fails to parse (e.g., `*Done = Recent` when label is `Done = Recent`)
2. Status values containing temporal keywords are misparsed (e.g., `10🔥Expedite = Now` has "Now" incorrectly parsed as a due date)

### `src/services/NaturalLanguageParser.ts`
- Strip status trigger character once before matching candidates
- Cleaner implementation with single-pass trigger stripping
- Maintains all existing functionality

### `tests/unit/services/NaturalLanguageParser.status-extraction.test.ts`
- Added comprehensive regression tests for trigger handling
- Added tests for temporal keywords in status values
- All 30 status extraction tests passing ✅

### `package-lock.json`
- Fixed version mismatch (was 3.25.6, now synced to 4.0.3 to match package.json)
- Removed outdated `peer: true` flags

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "tasknotes",
-	"version": "3.25.6",
+	"version": "4.0.3",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "tasknotes",
-			"version": "3.25.6",
+			"version": "4.0.3",
 			"license": "MIT",
 			"dependencies": {
 				"@codemirror/view": "^6.37.2",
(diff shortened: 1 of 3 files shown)


callumalpass:
Thank you for this @renatomen . I had noticed some strange behaviour with the status trigger in the past, but hadn't figured it out! I'll be merging this. 

I might make some changes afterwards just to bring the status trigger logic into line with the other custom triggers. 

Thanks again for your work here!

PR #1361, opened 2025-12-12
feat: Add option to disable inline task overlay for aliased links

diegomarzaa:
New user setting to control whether the interactive Task Link Overlay widget is rendered for links that utilize custom display text (aliases).

#### Motivation

When embedding tasks within prose (e.g., a meeting note sentence), the widget's inline presence can be disruptive or interfere with the flow of surrounding text.

This feature allows users to choose to hide the widget specifically for aliased links, effectively treating them as plain hyperlinks while keeping the widget active for non-aliased links.

*   `[[Task Name]]` (No alias, shows widget)
*   `[[Task Name|Check Status]]` (Alias present, hides widget)

#### Implementation Details

1.  **New Setting:** Added `disableOverlayOnAlias` (boolean, default `false`) to `TaskNotesSettings`.
2.  **Settings UI:** The new toggle is located in **Settings → Features → Inline Tasks**.
3.  **Rendering Logic:** Conditional checks were added to the primary link rendering processors:
    *   `src/editor/TaskLinkOverlay.ts` (Live Preview): Skips decoration if the link is a Wikilink containing a pipe (`|`).
    *   `src/editor/ReadingModeTaskLinkProcessor.ts` (Reading Mode): Uses a heuristic check to skip rendering if the link's displayed text does not match the original file path or the task title.

#### Review

*   The new setting is present in the UI.
*   When the setting is **disabled (default)**, aliased links still show the Task Overlay widget (current behavior preserved).
*   When the setting is **enabled**:
    *   Links with aliases (`[[Task|Alias]]`) do NOT render the widget in Live Preview or Reading Mode.
    *   Markdown links (`[Alias](path)`) *do* still render the widget for now.
    *   Links without aliases (`[[Task Name]]`) *do* still render the widget.


diegomarzaa:
Hi @callumalpass,

I saw the discussion on PR [#1117](https://github.com/callumalpass/tasknotes/pull/1117) regarding the concern that forcefully disabling overlays for aliased links would break existing user workflows.

This PR (#1361) provides the configured option that you requested.

1.  **Non-Breaking Change:** The feature is controlled by a new setting, **"Disable overlay for aliased links,"** which is **disabled by default**. This ensures that all existing users who rely on overlays with aliases see no change in behavior upon updating the plugin.
2.  **Addresses Navigation:** It fulfills the core need of preventing the widget from rendering on aliased links (e.g., those with block/section links: `[[Task#Section|Alias]]`), allowing them to function as plain navigation links.
3.  **Complete Solution:** My solution handles both Live Preview (`TaskLinkOverlay.ts`) and Reading Mode (`ReadingModeTaskLinkProcessor.ts`) correctly.

I believe this implementation meets your criteria for preserving user experience while adding the requested flexibility.

callumalpass:
Thanks a lot for this, @diegomarzaa . The implementation is nice--I'll be merging and including the changes in the next release.

diegomarzaa:
Great! Thanks!

PR #1400, opened 2025-12-24
fix: add mobile touch drag-and-drop for Kanban view

### LINKED ISSUE #1399: [Bug]: can't drag and drop in kanban view on mobile
### Bug Description

I'm using Obsidian 1.10.5 on latest Android on a Pixel with TaskNotes 4.1.3.

Using the kanban view, I can neither reorder the columns themselves (specifying their order doesn't work either), nor drag and drop tasks between them (grouped by status).

There is the oddest effect though - if I first long press on a task, and *then* drag the column it is in onto another, the task will move columns. 🤔

Portrait or landscape mode do not seem to affect this.

l-mb:
Drag and drop now works for both cards and columns on mobile (and still on the desktop).

Fixes #1399

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "tasknotes",
-	"version": "4.0.3",
+	"version": "4.1.3",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "tasknotes",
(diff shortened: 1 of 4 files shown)


callumalpass:
Thank you so much for this PR, @l-mb --this looks great! I'll be merging and including the changes in the next release.

PR #1402, opened 2025-12-25
Fix swimlane Kanban column grouping when Group By is a formula

bailob:
## **Description**
This PR fixes a glitch in the Bases Kanban swimlane mode when the view is grouped by a `formula.*` property and swimlanes are enabled (e.g. swimlane = Status).

## **Bug**
After editing a task’s metadata (e.g. changing `scheduled` so the formula result moves from “Overdue” to “Today”), the card would immediately jump into the `None` column in swimlane mode, and it would only correct itself after an Obsidian force reload or after changing the Bases sort (which forces a re-run).

Flat (non-swimlane) Kanban mode behaved correctly.

## **Root Cause**
In swimlane mode, the column key was re-derived from `pathToProps` (including cached `formula.*` outputs). After a frontmatter edit, Bases’ `groupedData` can update promptly, while cached formula outputs can lag, causing a temporary mismatch and a fallback to `None`.

## **Fix**
In `renderWithSwimLanes()`, column assignment now always uses the already-built `groups` map (which is derived from Bases `groupedData`). Swimlanes are then applied on top of that using the swimlane property value. This keeps swimlane mode consistent with flat mode and with Bases’ computed grouping.

## **How I tested**
- Obsidian desktop
- Bases Kanban view
- Group By: formula property (e.g. scheduled bucket formula)
- Swimlane: status
- Edited task `scheduled` so it changes buckets
- Verified the card stays in the correct column immediately (no reload / sort toggle needed)

## **Note**
This change was produced as a pure “vibe coding” patch and submitted via OpenAI Codex.

### DIFF
--- a/src/bases/KanbanView.ts
+++ b/src/bases/KanbanView.ts
@@ -8,7 +8,11 @@ import { createTaskCard } from "../ui/TaskCard";
 import { renderGroupTitle } from "./groupTitleRenderer";
 import { type LinkServices } from "../ui/renderers/linkRenderer";
 import { VirtualScroller } from "../utils/VirtualScroller";
-import { getDatePart, parseDateToUTC, createUTCDateFromLocalCalendarDate } from "../utils/dateUtils";
+import {
+	getDatePart,
+	parseDateToUTC,
+	createUTCDateFromLocalCalendarDate,
+} from "../utils/dateUtils";
 
 export class KanbanView extends BasesViewBase {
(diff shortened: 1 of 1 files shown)


callumalpass:
Thank you for this PR, @bailob . Very much appreciated!

callumalpass:
Merged manually via local merge and push to main. See commit 4201a523.

PR #1414, opened 2026-01-01
fix: map kanban priority labels in headers

normenmueller:
# fix-prio-label-mapping

## Kanban Priority Label Mapping

This change ensures Kanban column headers use the configured priority labels when grouped by priority (including swimlanes), instead of the raw stored values.

Examples (illustrative):

- Priority label display: a priority value like `high` with label "High Priority" renders as "High Priority" in the column header.

## Changelog

- Kanban column headers now map priority values to their configured labels when grouped by priority.

cf. [PR.md](PR.md)


callumalpass:
Thank you for this also, @normenmueller . This is a nice change! Much appreciated.

Issue #1391, opened 2025-12-20
[Bug]: Task identification by property does not identify the tasks correctly


loukandr:
### Bug Description

Task identification by property does not identify the tasks properly.

When creating tasks, it properly creates the chosen property name and value, but it still identifies only tasks with the task tag.

A restart of Obsidian is probably needed to reproduce the issue.

![Image](https://github.com/user-attachments/assets/52d24b5d-252b-4119-a95a-6208b98e2b14)

![Image](https://github.com/user-attachments/assets/4ea630ff-100f-4de1-83e4-1faac04ac96e)

ventoteso:
I can suggest a workaround:
1. Set Identify tasks by to "Tag".
2. In the Task tag field, make sure it is empty (remove any value). The default value is "task".
3. Set Identify tasks by back to Property.
This worked for me on v4.1.3.

loukandr:
> I can suggest a workaround:
> 
> 1. Set Identify tasks by to "Tag".
> 2. In the Task tag field, make sure it is empty (remove any value). The default value is "task".
> 3. Set Identify tasks by back to Property.
>    This worked for me on v4.1.3.

Tnx. This workaround seems to work until property identification becomes more stable.

callumalpass:
Thanks for reporting this @loukandr. I'll be looking into a fix for this.

PR #1416, opened 2026-01-02
fix: preserve tags in property-based identification (#1391)

normenmueller:
# fix-task-prop-based-ident

## Property-based task identification preserves tags

This change prevents property-based identification from mutating tags automatically. Tags are only written when the user explicitly sets them (including default tags), while the identifying property is still applied.

Examples (illustrative):

- Creating a task with property identification and tags like `client, urgent` keeps those tags; no extra task tag is injected.
- Updating priority/status without touching tags leaves existing tags unchanged.

### Task Identification

- Only add the identifying task tag to frontmatter when tag-based identification is enabled.
- In property-based mode, preserve user/default tags without filtering them out.
- Only write `frontmatter.tags` during updates when tags are explicitly provided; otherwise leave tags untouched.
- Fixes callumalpass/tasknotes#1391.

Cf. [PR.md](PR.md)

### DIFF
--- a/PR.md
+++ b/PR.md
@@ -22,3 +22,7 @@ Examples (illustrative):
 
 - Property-based identification now keeps all tags visible in the edit modal (only tag-based mode hides the task tag).
 - Tags are only updated when the user actually changes the tags field, preventing accidental tag removal on unrelated edits.
+
+## Risks / Notes
+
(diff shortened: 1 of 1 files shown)


callumalpass:
Thank you @normenmueller . This is a good fix! Merging.

PR #1443, opened 2026-01-07
fix: sync default status/priority when referenced value is deleted

l-mb:
When deleting the default value for a task status from the custom statuses, the `defaultTaskStatus` property would remain at the old value.

Since there's then no label associated with this value, the UI under Task Properties -> Status -> Default would show another label (the first one). So this looked as if it was "correct". Setting it explicitly to the intended default wouldn't update this either, since the UI filtered this as "no change".

The first time it was set to *another* value and then back would fix this.

This patch addresses this by ensuring that when the default status/priority is deleted, the default value is also updated. When the UI notices that it is set to a non-existing value, it is also updated.

### DIFF
--- a/src/settings/tabs/taskProperties/priorityPropertyCard.ts
+++ b/src/settings/tabs/taskProperties/priorityPropertyCard.ts
@@ -27,6 +27,13 @@ export function renderPriorityPropertyCard(
 		plugin.settings.fieldMapping.priority
 	);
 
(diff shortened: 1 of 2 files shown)


callumalpass:
Thank you for this PR, @l-mb. Sorry for the delay in reviewing it. Very happy to merge and include in the next release!

PR #1448, opened 2026-01-08
fix: persist project removals

normenmueller:
# fix/projects-removal

## Ensure project removals persist from the edit dialog

When removing project assignments in the task edit modal, the frontmatter now updates correctly (including fully removing the property when the list is emptied). We also normalize link comparisons so different link syntaxes (e.g., angle‑bracket markdown links) do not cause false change detection.

Examples (illustrative):

- Removing the last project now deletes the `projects` property from frontmatter.
- `[Project](<path/to/Project.md>)` and `[[path/to/Project]]` compare as the same target for change detection.

## Changelog

- Normalize project link comparisons in the edit modal to handle angle‑bracket markdown links.
- Persist empty project lists by explicitly removing the `projects` field from frontmatter.
- Add a unit test to ensure empty project updates remove the property.

## Tests

- `./node_modules/.bin/jest tests/unit/services/TaskService.test.ts --runInBand`

### DIFF
--- a/PR.md
+++ b/PR.md
@@ -1,0 +1,20 @@
+# fix/projects-removal
+
+## Ensure project removals persist from the edit dialog
+
+When removing project assignments in the task edit modal, the frontmatter now updates correctly (including fully removing the property when the list is emptied). We also normalize link comparisons so different link syntaxes (e.g., angle‑bracket markdown links) do not cause false change detection.
+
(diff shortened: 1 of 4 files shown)


callumalpass:
Thanks very much for this @normenmueller. I'll be merging.

PR #1458, opened 2026-01-09
feat: add css classes for context colors

phortx:
## Why
I'd love to give the context tags different colors via CSS. Unfortunately this is not possible currently as the elements are not distinguishable.

## What
I extended the `.context-tag` elements with a BEM modifier class (`context-tag--color-X`) that enables CSS-based coloring based on the context name.

That allows users to style their context tag elements. Also it is the very first step to bring configurable colored context elements as a feature in TaskNotes in the future.

## Changes
- Added `getContextColorClass()` function in `src/ui/renderers/tagRenderer.ts` that generates consistent color classes using a djb2 hash algorithm
- Context tags now receive a modifier class like `context-tag--color-0` through `context-tag--color-19` (so we get 20 distinct colors)
- Same context name always produces the same color class (case-insensitive, with/without @ prefix)
- Added unit tests

## Usage
CSS can now target specific color variants:

```css
.context-tag--color-0 { background-color: red; }
.context-tag--color-1 { background-color: blue; }
/* ... etc */
```

### DIFF
--- a/src/ui/renderers/tagRenderer.ts
+++ b/src/ui/renderers/tagRenderer.ts
@@ -74,8 +74,9 @@ export function renderContextsValue(
 	if (typeof value === "string") {
 		const normalized = normalizeContext(value);
 		if (normalized) {
+			const colorClass = getContextColorClass(normalized);
 			const el = container.createEl("span", {
-				cls: "context-tag",
+				cls: `context-tag ${colorClass}`,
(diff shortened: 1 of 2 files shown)


callumalpass:
Thanks for this PR, @phortx. Sorry it has taken a while to review. I'm happy to merge it and it will be included in the next release.

PR #1477, opened 2026-01-14
Add regex filter to .ics calendars

mcbabo:
Adds a regex filter to .ics calendars.

**Weird behavior:**
When updating the regex, the calendar view must be closed and then reopened to see the `ICSEntry` changes, I guess it's because of some caching (maybe `ICSCache`?)

-> Needs further investigation if wanted


mcbabo:
Any updates?

PR #1546, opened 2026-01-28
feat: keep kanban headers visible on scroll

normenmueller:


callumalpass:
Thank you @normenmueller! Merging!

Issue #701, opened 2025-09-19
[FR]: Task dependencies and sequencing within a project


prayidae:
### Feature Category

Task Management

### Problem Statement

The Projects functionality in Task Notes is elegant and very useful, and makes it easy to group related tasks together. But there is not currently a way to identify when a task strictly requires that another task be completed before it be started. 

This means it is not possible to filter in any of the task views for tasks which *it is currently possible to work on.* It's also not possible to filter for tasks that are in the way of other tasks. These are both valuable filter options that help with setting priorities, planning timeframes for projects, and figuring out where to get started. 

It's of course possible to create a custom user field/property key in a task note, `blockedBy`, and input other tasks as values of that property, and then filter for task note where that property is not empty. But the filter won't know whether the blocking task has been completed or not. In the workflow I am proposing, when one task is blocking another task and you mark the blocking task as completed, the blocked task will automatically be parsed by the plugin as *no longer blocked* (assuming it is not blocked by any other, uncompleted tasks.) 

This means that if you have a saved view of all tasks which it's currently possible to start work on, and you complete one of them that was blocking another task, the task it was blocking will automatically appear in the view. IE the next step in the project will immediately present itself to you. 

#### Core proposed feature

- Task notes have a property of type `list` called `blockedBy` (or `depends-on`, or `prerequisite`; the name doesn't matter)
  
- If it is not possible to begin Task B until Task A is complete, you can put a wikilink to Task A as a list item in the value of the `blockedBy` property of Task B. Task B is now considered "blocked by" Task A; Task A is considered "blocking" Task B
  
- In the Task List view and in other embedded views of TaskNotes bases (such as the subtasks base that displays in a project note), there are dedicated filter options for: 
	- **Blocked tasks:** Tasks where the value of property `blockedBy` contains a link to any *uncompleted* task note. 
	- **Unblocked tasks**: Tasks where the value of property `blockedBy` *does not contain* a link to any *uncompleted* task note. 
	- **Blocking tasks:** Tasks for which, somewhere in the vault, there is a note where property `blocked-by` contains this note

- It is possible to set a task as "blocking" or "blocked by" another task through either tasks's context menu and in the Edit Task Details modal. 

##### Sort base view by dependency chain: 
[...]

Issue #621, opened 2025-09-09
[FR]: Absolute order and support for reordering tasks with the same status in the Kanban view


dsebastien:
### Feature Category

Views (Calendar, Kanban, Agenda, Task List, Notes)

### Problem Statement

In the Kanban view, it would be nice to be able to define a "rank" property, which could then be used to store the position/rank of each task having the same status. This would enable defining an absolute order for tasks.

### Proposed Solution

Cfr above

### Use Case Examples

Drag and drop tasks to reoder them within the same column (state)

### Priority/Impact

Medium - This would be a nice improvement

callumalpass:
Thanks for the suggestion @dsebastien . I agree that this would be a nice feature to have. Closing as a duplicate of #386. 

I'm currently unsure how to approach this. The rank property could work, but I'm not sure how that property would interact with other views. 

For discussion of a related problem see #186

Issue #219, opened 2025-07-18
[Bug]: manual sorting in agenda task doesn’t work


Crunchnuggetz:
### What's the bug?

I tried to sort tasks in an order that went with my flow and it did not work. I believe this was an option in the last couple releases.

### What did you expect to happen?

Drag and drop in the task/agenda view would sort tasks manually 

### Which part of TaskNotes is affected?

None

callumalpass:
This has not been a feature in the past. I am working on figuring out an solution that feels right for the plugin. See #186

Crunchnuggetz:
Okay, thanks!

callumalpass:
Thanks for raising this! Closing this out — the Agenda view is date-ordered by design and doesn't support manual drag-reorder. For custom ordering, please use a Task List view sorted by Manual Order, which does support drag-to-reorder.

Issue #186, opened 2025-07-13
[FR]: Ordered Subtasks


Poly-0000:
### Feature Category

Task Management (creating, editing, organizing tasks)

### Problem Statement

For some of my projects, I like to have a particular order that my subtasks should be completed. It is currently not possible to have tasknotes display my subtasks in the order I want. the default sorting order is always used.

### Proposed Solution

I quite like [this idea proposed in the projects discussion](https://github.com/callumalpass/tasknotes/discussions/118#discussioncomment-13623313). Project tasks could have an additional frontmatter field that would determine whether the subtasks should be ordered. Additionally, an easy way to reorder tasks would need to be implemented. dragging tasks definitely be my preference.

### Use Case Examples

Most of my projects are not a lose collection of tasks, they are an ordered list of tasks that should be (but don't always have to be) completed in a specific order. here's a made up example:
```md
- [ ] gather ideas
- [ ] write an outline
- [ ] write a first draft
- [ ] get feedback
- [ ] write second draft
```
As I create new subtasks and plans change, I very often rearrange my subtasks, so I would very much prefer an easy way to rearrange them.

### Priority/Impact

Critical - I can't use TaskNotes effectively without this

callumalpass:
Thanks for this suggestion @Poly-0000 . This sounds like a good idea. I'll have to have a bit of a think about how to store the order data--for the moment this doesn't seem quite feasible with the current data structure. 

I'm curious if you have a work around for this already? I'd imagine the easiest solution would be to have the "Inline Widgets" (through the wikilinks) on a task note--and you'd rearrange these just as you rearrange any other text in a note. I feel like you've described [here](https://github.com/callumalpass/tasknotes/discussions/118#discussioncomment-13655544) some of the problems that might arise from this approach--is that right?

Poly-0000:
Hi, @callumalpass. I sadly haven't found an acceptable workaround for me so far. I've simply taken to focusing on my non-project tasks while I figure something out.

I really don't like the wikilink solution currently, as it makes creating tasks on the go a lot slower. It also comes with the same issues that you mentioned. additionally, it also just looks pretty jarring having the wikilinks in addition to the subtask widget.

callumalpass:
That makes sense. I should fairly easily be able to add the filter bar into the subtask widget, and this will provide some control over the order that the tasks appear in. I'll have to think more carefully, though, about how order information gets stored. There isn't an obvious solution for this, since the lists that appear are very dynamic. 

[...]

PR #1619, opened 2026-02-17
feat: add drag-to-reorder for kanban and task-list views

ac8318740:
## Summary

Adds drag-to-reorder for both kanban and task-list views using a `sort_order` frontmatter property.

The point of this is to let you prioritize tasks within a priority group, e.g. ordering your three "high" priority tasks in the sequence you actually want to do them. This is groundwork for auto-scheduling: once tasks have a stable intra-group ordering, a scheduler can slot them into time blocks without the user having to re-specify intent.

## What's included

- **Kanban view:** card-level drag-and-drop reordering within columns, midpoint insertion algorithm, collision handling (renumbers when neighbours share identical values)
- **Task-list view:** drag-to-reorder within groups, plus cross-group dragging (updates both `sort_order` and the group property)
- **Shared `sortOrderUtils.ts`:** extracted free functions (`computeSortOrder`, `getGroupTasks`, `renumberAndInsert`, etc.) so both views use the same algorithm
- **FieldMapper / types:** `sortOrder` field in `TaskInfo`, default field mapping, read/write support
- **CSS:** drop indicator pseudo-elements for visual feedback, simplified drag styling to fix hit-test issues
- **Configurable property name:** the frontmatter property used for sort order is configurable via Settings > Task Properties > Metadata Properties > Sort Order (defaults to `sort_order`)

## Configurable sort order property

The sort order property name is wired through the existing `FieldMapping` system, the same system used by every other property (status, priority, due, etc.). This means:

- **Default behaviour is unchanged** - the default mapping is `sort_order`, so existing users are unaffected
- **Users can customise it** in Settings > Task Properties, in the Metadata Properties section, via a "Sort Order" property card

## Design notes

- Midpoint insertion: `new_value = floor((above + below) / 2)` gives O(1) placement without rewriting the whole list
- Vault-wide neighbour lookup so filtered/hidden tasks are still considered (prevents collisions)
- Container-level drag event delegation so it works with both normal and virtual-scrolling rendering
- Deferred re-render during drag to prevent DOM destruction mid-drop

## Test plan

- [x] Existing unit tests pass (FieldMapper, bases)
- [x] Drag a card within a kanban column, verify order persists after reload
- [x] Drag a card in a grouped task-list, verify it stays in position
- [x] Drag a card between groups in task-list, verify the group property updates
- [x] Rapidly reorder many cards to trigger collision renumbering, verify no duplicate sort_order values
- [x] Change the sort order field mapping in settings to a custom name, confirm drag-to-reorder reads/writes the custom property
- [x] Confirm drag-to-reorder only activates when the view is sorted by the configured property name

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -37,15 +37,7 @@ Example:
 - Fixed a broken docs cross-reference from Property Types Reference to Task Properties settings
 - Fixed docs site link generation so internal Markdown links resolve to route URLs instead of broken `.md` paths (for example `/views/default-base-templates/`)
 - Fixed docs release-note links by building all Markdown docs pages, including pages not listed directly in sidebar nav
-- Drag-to-reorder now scopes Kanban ordering to the active swimlane, initializes sparse manual ordering using the visible drag order, rebalances oversized sort keys back to compact LexoRank values, and warns before large multi-note reorder writes
-- The default manual-order metadata property is now `tasknotes_manual_order`, and the setting is labeled "Manual Order" to make its drag-to-reorder role clearer
-- Task List drag-to-reorder now blocks formula-group drops and updates list-backed group properties safely during cross-group moves
-- Task List drag-to-reorder now computes persisted placement from the visible list order, reducing off-by-one drops when grouped sections or hidden tasks are involved
-- Fixed drag-to-reorder rank generation when a view sorts the manual-order field descending, so drops persist in the same direction the list is displayed
-- Grouped Task List drag previews now use a real boundary slot, so later group headers move out of the way instead of being overlapped during drag
-- Fixed Task List reorder mode swallowing subtask and dependency toggle clicks by treating interactive task-card controls as no-drag targets, with matching drag/pointer cursor feedback
-- Fixed Task List drag-to-reorder resolving against expanded subtask/dependency cards instead of the top-level rows, which could shift drops to the wrong task
-- Stabilized Task List drag-to-reorder hit-testing so the preview gap no longer shifts the effective drop boundary while you drag
-- Grouped Task List drag-to-reorder now resolves a Kanban-style insertion slot per group, so the previewed slot and the persisted drop stay aligned across group headers
-- Generated default Kanban, Relationships, and Tasks `.base` files now surface manual ordering by default where the view is task-focused, including a new grouped "Manual Order" task-list view
+- Added drag-to-reorder for Kanban and Task List views, including grouped Task List moves and safer handling for swimlanes, filtered views, sparse manual-order data, and large multi-note reorders
+- Improved drag-to-reorder feedback and interaction quality so drop previews stay aligned with the final result, grouped Task List sections behave more like Kanban, and interactive controls like subtask and dependency toggles remain clickable during reorder mode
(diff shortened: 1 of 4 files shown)


callumalpass:
Thanks so much for pushing this, @ac8318740 ! I’ve been interested drag-to-reorder in TaskNotes for a long time (see issues #186 , #219 , #621, #701) and hadn’t been brave enough to tackle it myself. This gets close and the direction is strong.

After testing in my vault, I’m seeing a few blocking robustness issues:

- When most notes don’t already have `sort_order`, reordering is pretty buggy and can land notes in unexpected places.
- Kanban drag/reorder in swimlane mode is unreliable
- There are still paths where writes look hardcoded to `sort_order`, which breaks custom mappings.

My suggestion for the next iteration: switch the ordering model to LexoRank (string ranks) and make DnD placement strictly swimlane-aware when swimlanes are enabled. This will require a lot of manual testing. I’d prioritize robustness over minimal complexity here, because this feature will get heavy use once merged.

Thank you for doing the hard part of getting this moving!

ac8318740:
Thanks so much for the review – really appreciate you getting to it so quickly. I've pushed an update that hopefully addresses everything you flagged:

- **LexoRank ordering** – switched from numeric midpoints to LexoRank strings, which handles the missing `sort_order` case properly now. Unranked neighbors get a rank assigned lazily via `ensureRank()` so subsequent drags stay consistent.
- **Swimlane-aware reordering** – `computeSortOrder` now threads swimlane params through so drags within a swimlane cell only consider tasks in that cell.
- **Custom field mapping** – threaded `sortOrderField` through all the internals instead of hardcoding `sort_order`. Should respect whatever's configured in settings now.
- **Robustness** – extracted a `rankBetween()` helper with try/catch around `LexoRank.between()` so equal-rank edge cases degrade gracefully instead of crashing. Also pinned the lexorank dep to exact version.

Also reworked the drop indicator UX while I was in there – replaced the thin accent line with an animated gap/slot pattern (cards shift apart to show where the drop will land), and fixed a couple of visual bugs with column resizing during drag.

ac8318740:
After further testing, I'm making updates/changes. Will follow-up with another commit here.

sookochoff:
Ability to reorder would be SUPER useful and has been sorely missing so far. Thank you for building this!

ac8318740:
Hey @callumalpass — just rebased onto main to clear the conflicts and wanted to flag this is ready for another look when you get a chance.

Since your review, quite a bit has changed:

- **LexoRank ordering** — switched from numeric midpoints to LexoRank strings. Unranked tasks get a rank lazily via `ensureRank()`, so the "most notes don't have sort_order" case is handled properly now. Also fixed edge cases with near-ceiling ranks and filtered views.
- **Swimlane-aware DnD** — `computeSortOrder` threads swimlane params through so reordering only considers tasks within the same cell. Also fixed the optimistic move path for swimlanes (wrong CSS selector meant cards were snapping back before the data update caught up).
- **Custom field mapping** — all sort order reads/writes go through `settings.fieldMapping.sortOrder`, no more hardcoded `sort_order` strings.
- **Cross-column snap-back fix** — cards now move instantly on drop (both swimlane and non-swimlane) instead of snapping back to source then jumping.
- **Debug logging** — gated behind a settings toggle so it's not polluting the console in normal use.

Happy to walk through anything or make changes if you have further feedback.

callumalpass:
Thanks very much for those changes @ac8318740 . I'll look into merging this (with a few edits) for the next release.

callumalpass:
Closing because this work has already been integrated onto main outside the PR branch ancestry. main now includes the reconciled drag-to-reorder work and follow-up fixes at commit 4ed824ad.

Issue #1109, opened 2025-11-13
[Bug]: Instant Task Convert breaks links


merchako:
### Bug Description

Suggestion: 
- [ ] When instantly converting a checklist item into a TaskNote
  - [ ] put links in the new TaskNote's body
  - [ ] keep the link text (if any) in the title

How to reproduce:
* Given a checklist item with a link in it,
* Convert that item to be a TaskNote,
* **Expectation**: the link will still exist somewhere
* **Observation**: the link has been sanitized and is no longer a link.

Example
<img width="752" height="118" alt="Image" src="https://github.com/user-attachments/assets/f0b9c2db-2b75-409e-af6f-16467402774d" />
<img width="762" height="122" alt="Image" src="https://github.com/user-attachments/assets/2d74e862-2482-401c-b999-b2177edfcb3e" />

```diff
- // Before
- - [ ] Alfred: Make quick way to change display resolution ([Resolution Changer](https://alfred.app/workflows/firefingers21/resolution-changer/))
+ // After
+ - [[Alfred Make quick way to change display resolution (Resolution Changer(httpsalfred.appworkflowsfirefingers21resolution-changer))-3]]
```

callumalpass:
Instant conversion now extracts markdown links and wikilinks from the converted title, keeps the display text in the task title, and appends the original links to the new task body under `Source links:`. That covers the requested behavior from this report and appears to have been fixed by the later link-preservation work around #1365. Focused tests around converted link preservation are passing. This looks resolved now. Thanks @merchako for the clear repro and expected behavior.

Issue #1102, opened 2025-11-11
[Bug]: Adjusting task status in frontmatter is not reflected in the taskcard


Wayment9504:
### Bug Description

When changing a task's status property in the frontmatter (ex. from todo to done or vice versa), the status of the task is not updated in the taskcard until I click on the task circle in the card.

When I right click on the task, the status in the pop-up menu reflects the correct status, but the taskcard itself does not.
This also means that recurring tasks do not trigger their rescheduling until I click on the taskcard itself to complete.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/38adf383-886b-4a82-82a0-16668b9172a4" />

callumalpass:
Thank you for reporting this. This should be resolved in release 4.8.0, so I am going to close this issue now.

PR #1631, opened 2026-02-20
fix: convert links from new task title to body #1109

eulucastr:
Fixing bug #1109


callumalpass:
Thanks @eulucastr , I'll take a look at this. I'll note that this seems to address issue #1109, rather than #1102.

PR #1664, opened 2026-03-04
Added projectId, project, and projects templates for filenames.

bendavis987:
Following the discussion on #1178, I have made changes that are less intrusive.

In the time that's passed, the code around these file name formats has gotten a lot clearer, so I've only added these as custom templates (filename only).

The projectId template essentially only shortens the project name to four characters in caps, and the filename exists check just increments this. 
- {{projectId}}-0 >> TASK-0, TASK-01. 

This seems to be the best option without combining the generateCustomFilename and generateUniqueFilename functions. This naming depends on the taskdata in custom and the vault access for the increment. Since this is handled by the custom naming, it also allows different combinations with this projectId. Like my new setting below.
- {{shortYear}}Q{{q}}{{projectId}}-{{title_snake}} >> 26Q1TASK-my_task.

Would you be interested in unifying these replacements? IE: folderTemplateProcessor is essentially duplicated, but not as clear. I'd be willing to look into this.

### DIFF
--- a/docs/settings/task-defaults.md
+++ b/docs/settings/task-defaults.md
@@ -24,6 +24,7 @@ The **Default Tasks Folder** setting supports dynamic folder creation using temp
 - `{{contexts}}` - All contexts from the task's contexts array, joined by `/`
 - `{{project}}` - First project from the task's projects array  
 - `{{projects}}` - All projects from the task's projects array, joined by `/`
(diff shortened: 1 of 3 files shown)


rdezwart:
Love these changes, thank you! The documentation mentions body templates accepting all filename template variables, however these project variables do not seem to be included. I'm not familiar with the codebase, would this require separate implementation?

PR #1672, opened 2026-03-05
fix: add "custom" to MCP time summary period enum

pib:
## Summary

The `tasknotes_get_time_summary` MCP tool declares `from` and `to` parameters for custom date ranges, but they never work because the Zod enum for `period` only allows `today|week|month|all`. The HTTP API (`/api/time/summary`) works fine with `period=custom&from=...&to=...` because it has no enum validation.

## Root cause

`src/services/MCPService.ts` line 529 — Zod enum validation prevents any period value other than `today`/`week`/`month`/`all` from reaching the `default` case in `computeDateRange()`, which is where `fromDate`/`toDate` are used.

## Fix

Add `"custom"` to the Zod enum. The `default` case in `computeDateRange` already handles the rest correctly.

## Test plan

- [x] MCP call with `period=custom`, `from`, `to` returns data scoped to custom range
- [x] New unit test for `computeTimeSummary` with custom period and date range filtering
- [x] All 21 time tracking tests pass (20 existing + 1 new)
- [x] Existing period values (`today`, `week`, `month`, `all`) unaffected
- [x] Custom period without `from`/`to` falls back to today (no error)
- [x] Invalid period value correctly rejected by Zod validation
- [x] Local plugin build + install verified in live Obsidian vault

### DIFF
--- a/tests/unit/utils/timeTrackingUtils.test.ts
+++ b/tests/unit/utils/timeTrackingUtils.test.ts
@@ -260,6 +260,40 @@ describe('timeTrackingUtils', () => {
 			expect(result.topTags).toBeUndefined();
 		});
 
+		it('should filter by custom date range when period is "custom"', () => {
+			const tasks: TaskInfo[] = [
+				TaskFactory.createTask({
+					title: 'In Range',
+					path: '/tasks/in-range.md',
+					timeEntries: [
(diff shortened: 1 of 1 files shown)


callumalpass:
Thanks very much for this @pib! I'll be merging this for the next release.

PR #1679, opened 2026-03-06
Open the timeblock modal with the task title prefilled from the task context menu

### LINKED ISSUE #1678: [FR]: Open the timeblock modal with the task title prefilled from the task context menu
### Feature Description

When planning in Calendar view, creating a timeblock from a task currently requires manually retyping the task title. This adds friction to weekly planning workflows.

We should add a “Create timeblock” action from the task context menu that opens the timeblock modal with:

- title
- sensible default date/time based on task scheduling info (or fallback defaults)

Users should still be able to edit the prefilled title and times before saving.

## Why

- Reduces repetitive typing
- Makes task-to-timeblock planning faster
- Keeps task and timeblock naming consistent by default while preserving flexibility

## Expected behavior

- Right-click task -> select Create timeblock
- Timeblock modal opens with task title prefilled
- Date/time defaults:
  - Use task scheduled datetime when available
  - Otherwise use selected target date and a default planning slot
  - End time uses task estimate when available, otherwise default duration
- User can modify any prefilled value
- Saving creates a normal timeblock entry

Lorite:
Fixes #1678
- Open the timeblock modal with the task title prefilled from the task context menu
- In Create and Edit Timeblock, adding an attachment now auto-fills the title when the title is empty
- Add setting for attachment search order in timeblock modal
- Add an "Add task" action to Create Timeblock and Edit Timeblock to select a task and prefill the title

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -28,7 +28,7 @@ Example:
 
 - Added a task context-menu action to create a timeblock with the task title prefilled, and default times derived from the task schedule or a planning slot
 - Timeblocks created from a task now automatically include that task note as a prefilled attachment
-- In Edit Timeblock, adding an attachment now auto-fills the title when the title is empty
+- In Create Timeblock and Edit Timeblock, adding an attachment now auto-fills the title when the title is empty
 - Added a setting to control the ordering of results in the timeblock Add Attachment search window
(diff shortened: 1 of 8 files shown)


cosmic-lighthouse:
100% think we should talk to Things 3 about an optional clone base, let me know what you think 🤙

PR #1683, opened 2026-03-07
docs: fix sidebar overlay being always present on smallest media query

pierosavi:
Currently there's a bug in the docs at https://tasknotes.dev where the smallest media query `@media (max-width: 720px)` sets the sidebar overlay to `display: block` on top of anything else, making the content impossible to interact with. 
I corrected the css to set the expected behavior.

### DIFF
--- a/docs-builder/src/styles/main.css
+++ b/docs-builder/src/styles/main.css
@@ -856,10 +856,11 @@ body {
   }
 
   .overlay {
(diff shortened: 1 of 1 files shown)


callumalpass:
Thanks for catching this and for the fix @pierosavi ! I'll be merging :)

PR #1687, opened 2026-03-08
Fix a bug about resetting the calendar view to today incorrectly

### LINKED ISSUE #1513: [Bug]: Calendar jumps back to current week when clicking a non-current week number in Year view
### Bug Description

**Description**
When I'm in Year view and click on a week number that is not the current week, the calendar briefly navigates to the correct week but then automatically jumps back to the current week. This makes it impossible to stay on and inspect details of other weeks.

**Expected Behavior**
After clicking any week number in Year view, the calendar should stay on that week and not revert to the current week unless the user hits 'Today'.

**Environment**
Obsidian version: 1.11.5
TaskNotes plugin version: 4.3.0

**Steps to Reproduce**
Open a calendar view
Switch to Year view
Click on any week number that is not the current week
Observe that the calendar briefly shows the selected week, then jumps back to the current week

### LINKED ISSUE #1686: [Bug]: When creating a new timeblock from the advanced calendar view in a different week, the view resets to the current week
### Bug Description

When creating a new timeblock from the advanced calendar view in a different week, the view resets to the current week.

Also related to #1513

Lorite:
Fixes #1513
Fixes #1686

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -26,6 +26,7 @@ Example:
 
 ## Fixed
 
+- Fixed a calendar view issue where creating or modifying events, timeblocks, etc. could reset the visible date back to today
 - Fixed documentation deployment CI failures caused by `docs-builder/src/js/main.js` being excluded by a broad `.gitignore` `main.js` rule
   - Added a specific unignore rule so the docs site client script is tracked and available in GitHub Actions builds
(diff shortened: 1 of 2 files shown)


callumalpass:
Closing because this work has been integrated onto main via merge commit cfe9df1d, including the follow-up fix to preserve the visible date only for incidental calendar recreates.

callumalpass:
Thanks very much for this fix, @Lorite . This will be included in the next release

PR #1706, opened 2026-03-16
Fix Google Calendar archive cleanup reliability

### LINKED ISSUE #1695: [Bug]: Archived tasks can remain on Google Calendar and orphan their event ID
### Bug Description

I hit an issue with Google Calendar export + auto-archive.

## Summary

When a scheduled task is marked `done` and later auto-archived, the TaskNote can be moved into `TaskNotes/Archive` but its Google Calendar event can remain visible.

In some cases the archived note still contains `googleCalendarEventId`. In worse cases, the note loses `googleCalendarEventId` even though the Google event still exists, which makes the remote event effectively orphaned because there is no local event ID left for retry/cleanup.

## Environment

- TaskNotes 4.4.0
- Obsidian desktop on macOS
- Google Calendar export enabled
- `syncOnTaskComplete: true`
- `syncOnTaskDelete: true`
- `syncTrigger: scheduled`
- `done` status configured with `autoArchive: true`

## Steps to reproduce

1. Enable Google Calendar export.
2. Create a scheduled task so it syncs to Google Calendar and receives a `googleCalendarEventId`.
3. Mark the task `done`.
4. Wait for auto-archive to move it into `TaskNotes/Archive`.
5. Check Google Calendar and the archived note frontmatter.

## Expected behavior

- Once the task is archived, the Google Calendar event should be deleted.
- `googleCalendarEventId` should only be removed from the note after the delete succeeds, or if Google returns `404`/`410` because the event is already gone.

## Actual behavior

- The task can be archived successfully while the Google Calendar event remains visible.
- Sometimes the archived note still contains `googleCalendarEventId`.
- Sometimes `googleCalendarEventId` is removed even though the Google event still exists, so the local pointer needed to retry deletion is lost.

## Suspected cause

There seem to be two related issues in the current implementation.

### 1. Archive-time delete is fire-and-forget

In `src/services/TaskService.ts` around lines `1086-1093`, `toggleArchive()` does this:

- checks whether the task is being archived
- calls `deleteTaskFromCalendar(updatedTask).catch(...)`
- does **not** await it

That means the archive operation can succeed even if calendar deletion fails or never completes.

### 2. Event ID is always cleared even after real delete failures

In `src/services/TaskCalendarSyncService.ts` around lines `919-945`, `deleteTaskFromCalendar()`:

- attempts to delete the Google event
- logs non-`404`/`410` failures
- then still calls `removeTaskEventId(task.path)` unconditionally

That means a real delete failure can still clear the local `googleCalendarEventId`, leaving the Google event behind with no local reference to clean it up later.

## Why this matters

This creates a bad failure mode:

1. task is marked `done`
2. auto-archive runs
3. task moves into `TaskNotes/Archive`
4. Google Calendar event is not deleted
5. local `googleCalendarEventId` may already be gone

At that point the archive state and calendar state are inconsistent.

## Suggested fix

I think the archive path should be made transactional / retry-safe:

- await archive-time `deleteTaskFromCalendar()` in `toggleArchive()`
- only remove `googleCalendarEventId` after:
  - successful delete, or
  - `404` / `410` from Google
- if needed, preserve the pre-archive event ID across the archive move so the delete step still has the ID even if refreshed task data drops it

## Additional note

I tested a local patch that:
- awaited archive-time deletion
- retried deletion
- avoided clearing `googleCalendarEventId` after arbitrary delete failures

That resolved the issue in my local testing.

Happy to test a fix if helpful.

martin-forge:
Fixes #1695.

## What this changes

- await Google Calendar cleanup during `toggleArchive()` instead of firing it and forgetting it
- make `deleteTaskFromCalendar()` return success/failure and only clear `googleCalendarEventId` after a confirmed delete or a `404`/`410`
- preserve the pre-archive event ID across archive moves so deletion still has the identifier it needs
- keep auto-archive queue items pending when the task is archived locally but Google cleanup still has to be retried
- retry Google cleanup for already-archived tasks that still have a lingering Google Calendar link

## Why

The previous flow could archive a task locally while the Google event remained behind, and on some failure paths the local `googleCalendarEventId` could be cleared anyway. That leaves an orphaned event with no local pointer for retry.

This PR makes the archive flow retry-safe: local archive can still succeed, but the event ID is preserved until Google deletion actually succeeds, and the auto-archive queue keeps retrying cleanup while the link remains.

## Tests

- `npx jest tests/unit/services/TaskService.test.ts --runInBand`
- `npx jest tests/unit/issues/issue-google-calendar-archive-reliability.test.ts --runInBand`

I also validated the patch on a clean upstream copy after generating the release notes bundle required for typecheck/build in this repo.

### DIFF
--- a/src/services/TaskService.ts
+++ b/src/services/TaskService.ts
@@ -69,28 +69,17 @@ export class TaskService {
 		task.googleCalendarEventId = undefined;
 	}
 
-	private async deleteArchivedTaskFromCalendarWithRetry(
-		task: TaskInfo,
-		maxAttempts = 3
-	): Promise<boolean> {
+	private async deleteArchivedTaskFromCalendar(task: TaskInfo): Promise<boolean> {
 		if (!this.plugin.taskCalendarSyncService) {
 			return true;
 		}
 
-		for (let attempt = 1; attempt <= maxAttempts; attempt++) {
-			const deleted = await this.plugin.taskCalendarSyncService.deleteTaskFromCalendar(task);
-			if (deleted) {
-				return true;
-			}
-
-			if (attempt < maxAttempts) {
-				// Brief backoff gives transient API failures a chance to clear
-				// before we fall back to the longer auto-archive retry cycle.
-				await new Promise((resolve) => setTimeout(resolve, attempt * 500));
-			}
+		const deleted = await this.plugin.taskCalendarSyncService.deleteTaskFromCalendar(task);
+		if (deleted) {
+			return true;
 		}
 
-		console.warn("Failed to delete archived task from Google Calendar after retries:", {
+		console.warn("Failed to delete archived task from Google Calendar during archive:", {
 			taskPath: task.path,
 			eventId: task.googleCalendarEventId,
 		});
@@ -1096,7 +1085,7 @@ export class TaskService {
 					updatedTask
 				);
 				archiveCalendarCleanupComplete =
-					await this.deleteArchivedTaskFromCalendarWithRetry(archiveCalendarTask);
+					await this.deleteArchivedTaskFromCalendar(archiveCalendarTask);
 				if (archiveCalendarCleanupComplete) {
 					this.clearGoogleCalendarMetadata(updatedTask);
 				}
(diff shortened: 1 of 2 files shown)


callumalpass:
Thanks very much for this pr, @martin-forge ! Merging for the next release

PR #1788, opened 2026-04-11
fix: restore subgroup indentation in TaskList views

renatomen:
## Description

Fixes subgroup headers not displaying with proper indentation in TaskList views when sub-grouping is configured.

## Problem

When viewing tasks with both primary grouping and sub-grouping enabled, sub-headers were rendering at the same level as primary headers without visual indentation, making the hierarchy unclear.

**Before:** Sub-headers appeared flush left with primary headers  
**After:** Sub-headers display with 16px left padding showing clear hierarchy

## Solution

Added CSS rules for `[data-level="sub"]` attribute selector in `task-list-view.css`:
- 16px left padding for indentation
- Smaller font size (0.95em) and muted color for visual distinction  
- Reduced opacity and size for chevron icons on sub-headers

The TypeScript code was already setting `data-level="sub"` on sub-header elements, but the corresponding CSS rules only existed in the Bases-specific stylesheet. This PR adds the missing rules to the general TaskList view styles.


callumalpass:
Hey @renatomen ! Hope you've been keeping well. Thanks for this--merging for the next release :)

Issue #1751, opened 2026-04-01
[FR]: Hide standalone Subtasks, if Parent / Subtask pair is already present


Glint-Eye:
### Feature Description

Now that the "expanded relationships: inherit" option allows to also apply filter settings to subtasks, there should also be a setting that hides Standalone Subtasks, that are already present in a Parent-Task / Subtask pair. Right now if filter settings apply to parent task and subtask a task list will show an expandable parent task with indented subtask and then additionally another instance of the subtask as a standalone task (so not indented under the parent task). This is redundant and only adds noise to a filtered list.

Proxybat:
Agree, I'd also like an easy option to hide standalone subtasks as I use the nested main tasks for organization. 

It's easier for me to keep track of things. With the current update it now filters the subtasks that were nested making the filter I use for hiding subtasks `projects.isEmpty()` outside nested main tasks useless. Test this on a sandbox vault in case it was a bug on my vault & I still have this issue. If anyone would like to suggest a workaround until a toggle/option to hide standalone subtasks gets implemented I'd really appreciate it. Screenshots for clarity:

<img width="1860" height="433" alt="Image" src="https://github.com/user-attachments/assets/0c51f645-76d1-4d1f-8c30-62bec31f8b95" />

<img width="1866" height="376" alt="Image" src="https://github.com/user-attachments/assets/9f4a247b-6814-4bb2-b86f-a244daa65959" />

callumalpass:
Resolved in 4.10.0. Thanks @Glint-Eye for opening this issue. This issue is being closed.

PR #1792, opened 2026-04-12
feat: add collapseSubtasks option to hide subtasks from top-level views

stanley-910:
## Summary

- Adds a `collapseSubtasks` view option for TaskList and Kanban views that hides subtasks from the top-level when their parent task is also present in the current result set
- Subtasks remain accessible via the parent task's expansion chevron
- Tasks linked to non-task project notes (e.g., a project hub page) are **not** filtered out, since the project note won't be in the task result set
- Automatically forces `expandedRelationshipFilterMode` to `show-all` when enabled, so expanded subtasks still display correctly
- Adds a "Collapse subtasks into parent" toggle to the Configure View menu for both view types

Addresses community request in #1751 — subtasks that appear both as expanded children under a parent and as standalone top-level entries create visual noise and confusion in filtered views.

### Usage

Add `collapseSubtasks: true` to a view's options in the `.base` file, or toggle "Collapse subtasks into parent" in the Configure View menu.

```yaml
views:
  - type: tasknotesTaskList
    name: Tasks
    collapseSubtasks: true
```

### Files changed

- `src/bases/BasesViewBase.ts` — shared `filterSubtasksFromTopLevel()` method that resolves each task's `projects` links and checks if any point to another task in the current result set
- `src/bases/TaskListView.ts` — reads `collapseSubtasks` config, applies filter before rendering
- `src/bases/KanbanView.ts` — same treatment
- `src/bases/registration.ts` — registers the toggle in the Configure View menu for both view types
- `tests/unit/bases/collapseSubtasks.test.ts` — 11 unit tests covering core behavior, edge cases (broken links, plain text, multi-project tasks, nested subtasks, parent not in view)

## Test plan

- [x] Unit tests pass (`npx jest tests/unit/bases/collapseSubtasks.test.ts` — 11/11)
- [x] Build succeeds with no TypeScript errors
- [ ] Manual: enable `collapseSubtasks: true` on a TaskList view — subtasks of visible parents should disappear from top-level
- [ ] Manual: expand a parent task's chevron — subtasks should still appear nested
- [ ] Manual: tasks linked to non-task project notes should remain visible at top-level
- [ ] Manual: toggle "Collapse subtasks into parent" in Configure View menu persists correctly

### DIFF
--- a/src/bases/BasesViewBase.ts
+++ b/src/bases/BasesViewBase.ts
@@ -10,6 +10,7 @@ import { TaskSearchFilter } from "./TaskSearchFilter";
 import { BatchContextMenu } from "../components/BatchContextMenu";
 import type { TaskCardOptions } from "../ui/TaskCard";
 import { BasesConfigLike, BasesQueryResultLike } from "./types";
+import { parseLinkToPath } from "../utils/linkUtils";
 
 /**
  * Abstract base class for all TaskNotes Bases views.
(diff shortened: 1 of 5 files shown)


callumalpass:
Thanks very much for this PR @stanley-910 . Inspired by this work, there's a similar fix for this issue in 4.10.0. Appreciate the thought put into this PR

PR #1802, opened 2026-04-16
Fix recurring exception sync for rescheduled recurring tasks

### LINKED ISSUE #1696: [Bug]: Google Calendar export ignores independently rescheduled next occurrence for recurring tasks
## Summary

TaskNotes 4.4.0 appears to support rescheduling the next occurrence of a recurring task without changing the long-term recurrence pattern.

However, Google Calendar export does not appear to honor that distinction. The next occurrence can be rescheduled inside TaskNotes, but the Google Calendar event remains anchored to the recurring pattern date.

## Why I think this is intended TaskNotes behavior

The official docs describe recurring tasks as having:

- a recurring pattern controlled by `DTSTART` / `recurrence`
- a next occurrence controlled by `scheduled`

The docs also describe dragging the solid-border recurring event to reschedule the next occurrence without changing the recurrence pattern.

In the installed 4.4.0 plugin bundle, I can also see the notice text:

`Rescheduled next occurrence. This does not change the recurrence pattern.`

So this seems like an existing feature in TaskNotes itself, not a requested new capability.

## Steps to reproduce

1. Create a recurring task with a stable recurrence pattern.
2. Example initial state:
   - `scheduled: 2026-03-13`
   - `recurrence: DTSTART:20260313;FREQ=WEEKLY;INTERVAL=4;BYDAY=FR`
   - `recurrence_anchor: scheduled`
3. Enable Google Calendar export.
4. Reschedule only the next occurrence inside TaskNotes, either by:
   - dragging the solid-border recurring event in the TaskNotes calendar, or
   - editing `scheduled` to a later date while intentionally keeping the recurrence pattern unchanged
5. Example resulting state:
   - `scheduled: 2026-03-16`
   - `recurrence: DTSTART:20260313;FREQ=WEEKLY;INTERVAL=4;BYDAY=FR`
6. Check Google Calendar.

## Expected behavior

Google Calendar should reflect the rescheduled next occurrence without permanently shifting the whole recurring series.

In other words:
- the next occurrence should move to the new scheduled date
- future recurring instances should remain anchored to the original recurrence pattern

I do not mind how this is implemented internally:
- recurrence override / exception
- separate one-off event plus skipped original occurrence
- any other equivalent model

But the exported calendar should match TaskNotes' intended recurring-task behavior.

## Actual behavior

Google Calendar remains anchored to the recurring pattern date instead of the independently rescheduled next occurrence.

In practice, TaskNotes and Google Calendar diverge:
- TaskNotes shows the next occurrence on the rescheduled date
- Google Calendar still shows the old recurrence date

## Suspected cause

The Google Calendar exporter appears to serialize recurring tasks primarily from the RRULE / `DTSTART`, and does not preserve the separate “next scheduled occurrence” concept that TaskNotes supports internally.

So the issue seems to be specifically in Google Calendar export, not in the core recurring-task UI model.

## Why this matters

This makes recurring tasks unreliable for users who:
- use TaskNotes as the source of truth
- sync scheduled tasks to Google Calendar
- sometimes postpone only the next occurrence while keeping the long-term cadence unchanged

That is a very normal recurring-task workflow.

## Request

Please make Google Calendar export honor independently rescheduled next occurrences for recurring tasks.

martin-forge:
## Maintainer note
Ready to merge. This is distinct from #1843: it fixes recurring task exception state and detached exception events. The only overlap is calendar cleanup. If #1843 lands first, this PR should be rebased so detached exception cleanup uses #1843's retry-safe deletion queue. If this lands first, #1843 should be rebased so its persisted event index/recovery also tracks detached exception event IDs.

## Summary
- Preserve the original RRULE occurrence date when a scheduled-anchor recurring task is moved off-pattern.
- Sync moved occurrences to Google Calendar as a recurring master with an EXDATE plus one detached exception event.
- Clean up detached exception metadata/events when the moved occurrence is completed, skipped, archived, deleted, or unlinked.
- Keep local calendar recurring rendering aligned by hiding original pattern dates represented by moved-occurrence metadata.

Closes #1696.

## Validation
- TZ=UTC npm test -- --runInBand
- npm run docs:check
- npm run i18n:sync
- npm run lint
- npm run typecheck
- npm run build:test
- obsidian plugin:reload id=tasknotes vault=test


martin-forge:
Cross-PR compatibility note from auditing against #1843:

This PR remains a distinct fix: #1843 does not add the recurring-exception metadata, moved-original-date tracking, local calendar rendering changes, or detached Google exception event flow introduced here.

The only substantive overlap is cleanup. If #1843 lands before this PR, please route this PR's primary + detached exception event cleanup through #1843's retry-safe deletion queue and keep 404/410 deletes as success. If this PR lands first, #1843 should be rebased so its persisted event index and startup recovery track both `googleCalendarEventId` and `googleCalendarExceptionEventId`; otherwise a task file deleted while Obsidian is closed could still leave a detached recurring exception event orphaned.

martin-forge:
Superseded by #1851, which consolidates the Google Calendar fixes into one ready-to-review PR.

PR #1810, opened 2026-04-19
fix(i18n): Update Japanese translations for Kanban terms

sigma7863:
### Summary

In the Japanese locale, "kanban" was translated as "かんばん" (hiragana). This PR updates it to "カンバン" (katakana), which is more commonly used for the term in a software / IT context.

### Changes

* Replace "かんばん" with "カンバン"

### Rationale

In Japanese, "kanban" as an IT / productivity term is typically written in katakana ("カンバン"). The hiragana form feels less natural in this context.

### Notes

Checked for consistency with other translations and terminology.

### DIFF
--- a/src/i18n/resources/ja.ts
+++ b/src/i18n/resources/ja.ts
@@ -282,7 +282,7 @@ export const ja: TranslationTree = {
 			},
 		},
 		kanban: {
-			title: "かんばん",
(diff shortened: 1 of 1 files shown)


callumalpass:
Thanks very much for this @sigma7863 . I'll be merging for the next release.

PR #1832, opened 2026-04-26
Normalize zero-duration external calendar events

### LINKED ISSUE #1823: [Bug]: Google Calendar zero-duration recurring exception can render on two days in list view
### Bug Description

I’ve found another Google Calendar rendering bug, but this one looks distinct from the sync/export issues.

A Google Calendar event can show up on **two different days** in TaskNotes list/agenda-style calendar views, even though it only exists once in Google Calendar.

What seems to trigger it is:

- a **recurring Google Calendar event**
- one instance being **moved off its original series date**
- the moved instance being a **timed event with `start == end`** (zero-duration / point-in-time event)
- the moved time being **late in the day**

I hit this with a moved recurring Google event at `23:12`, and TaskNotes rendered the same event once on the correct day and again on the following day.

This does **not** appear to be duplicate data coming from Google. I checked the Google Calendar event itself and there was only one instance in the relevant window.

### Steps to Reproduce

1. Use the Google Calendar integration.
2. Have a recurring Google Calendar event.
3. Move one occurrence off its original series date to a different day.
4. Make that moved occurrence a timed event where the start and end are the same (for example `23:12` to `23:12`).
5. Open a TaskNotes calendar view in `listWeek` / agenda-style mode.
6. Refresh calendars.

### Expected Behavior

The event should appear **once**, on the actual day/time returned by Google Calendar.

### Actual Behavior

The same Google Calendar event is rendered under **two day headers**: the correct day and the following day.

### Why I think this is in TaskNotes, not Google

I checked the event through the Google Calendar connector and there was only one matching event in the time window.

The Google event had a real `start` and `end`, but they were equal:

- `start: 2026-04-22T23:12:00+01:00`
- `end: 2026-04-22T23:12:00+01:00`

From the installed bundle / source, it looks like:

- `GoogleCalendarService.convertToICSEvent()` preserves that as a timed event with identical `start` / `end`
- `createICSEvent()` passes it through to FullCalendar unchanged
- the duplicate happens at render time in the calendar/list view, not during fetch

So this looks like a rendering/normalisation bug around **zero-duration timed Google recurring exceptions**, not a duplicate-fetch bug.

### Related but Distinct

This seems adjacent to the recurring exception work in #1696 / #1802, but it is not the same problem.

- #1696 is about exporting moved recurring occurrences correctly
- this one is about **displaying inbound Google Calendar events correctly** in TaskNotes calendar/list views

### Suggestion

I think the safest fix is probably to normalise zero-duration timed external events before handing them to FullCalendar, rather than letting them through as `start == end`.

martin-forge:
## Maintainer note
Ready to merge independently. This PR only normalizes zero-duration external calendar event rendering and does not depend on #1802, #1843, or #1844.

## Summary
- Clamp zero-duration timed external calendar events to a minimal positive display duration before passing them to FullCalendar.
- Preserve the original provider event data in `extendedProps` for context menus and debugging.
- Preserve UTC and explicit-offset formatting when normalizing zero-duration ranges, while leaving all-day and non-zero timed events unchanged.

Closes #1823.

## Notes
This PR is independent of #1802 and is based on current upstream `main`.

## Validation
- TZ=UTC npm test -- tests/unit/issues/issue-1823-zero-duration-google-calendar-list-duplication.test.ts --runInBand
- npm run docs:check
- npm run i18n:sync
- npm run lint
- npm run typecheck
- TZ=UTC npm test -- --runInBand
- npm run build:test
- obsidian plugin:reload id=tasknotes vault=test

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -23,3 +23,10 @@ Example:
 ```
 
 -->
+
+## Fixed
+
+- (#1823) Fixed zero-duration timed external calendar events rendering on multiple days in list-style calendar views
+  - Adds a minimal display duration before passing point-in-time external events to FullCalendar
+  - Preserves the original provider event data for context menus and debugging
+  - Thanks to @martin-forge for reporting and debugging
--- a/src/bases/calendar-core.ts
+++ b/src/bases/calendar-core.ts
@@ -578,11 +578,17 @@ export function createICSEvent(icsEvent: ICSEvent, plugin: TaskNotesPlugin): Cal
 			subscriptionName = subscription.name;
 		}
 
+		const { start, end } = normalizeExternalTimedEventRange(
+			icsEvent.start,
+			icsEvent.end,
+			icsEvent.allDay
+		);
+
 		return {
 			id: icsEvent.id,
 			title: icsEvent.title,
-			start: icsEvent.start,
-			end: icsEvent.end,
+			start: start,
+			end: end,
 			allDay: icsEvent.allDay,
 			backgroundColor: backgroundColor,
 			borderColor: borderColor,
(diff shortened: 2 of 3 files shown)


martin-forge:
Superseded by #1851, which consolidates the Google Calendar fixes into one ready-to-review PR.

PR #1838, opened 2026-04-28
fix: strip time component in default base template date comparisons

### LINKED ISSUE #1124: [Bug]: 4.0.0-beta 3 Today's scheduled tasks not showing
### Bug Description

When a task is scheduled for 14:00 today, it is not shown when the filter is 'scheduled <= today()'

But at 14:00, it is shown when the filter is `scheduled <= today() + "14 hours"`

I am in AEDT.

loukandr:
## Problem

`today()` returns midnight in the Bases formula language, so day-level comparisons in `tasks-default.base` fail for any value with a time component. A task scheduled for today at a non-midnight time (e.g. `15:45`) is excluded from the Today view because `date(scheduled) == today()` evaluates `<today> 15:45 == <today> 00:00` → false. The same issue affects `>= today()` and `<= today() + "Nd"` checks across several formulas.

Reported in #1124.

## Repro

1. Create a task with `scheduled: 2027-01-15T15:45` (use today's date, any non-midnight time).
2. Open the Today view in `tasks-default.base`.
3. Task is missing.

## Fix

Add `.date()` to strip the time component before comparing, matching the `isDueToday` and `isScheduledToday` formulas in the same file.

Affected:
- View filters: Today, This Week
- Formulas: `isDueThisWeek`, `isThisWeek`, `dueDateCategory`, `nextDateCategory`, `dueDateDisplay`

`< today()` comparisons (Overdue branches) are left as-is, they evaluate correctly without `.date()`.

## Tests

Extended `tests/unit/templates/defaultBasesFiles.test.ts` with positive pins for the affected formulas and negative matchers against the time-naive shapes.

Fixes #1124

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -23,3 +23,12 @@ Example:
 ```
 
 -->
+
+## Fixed
+
+- (#1124) Fixed `tasks-default.base` views and formulas excluding tasks scheduled or due today at a non-midnight time from day-level comparisons
+  - `today()` returns midnight, so `date(due) == today()` and `date(due) <= today() + "7d"` evaluated to false for any value with a non-zero time
+  - Affected views: Today, This Week
+  - Affected formulas: `isDueThisWeek`, `isThisWeek`, `dueDateCategory`, `nextDateCategory`, `dueDateDisplay`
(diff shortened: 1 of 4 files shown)


callumalpass:
Merged this fix via #1881 after resolving it against the current main stack and preserving the release-note acknowledgement. Thanks @loukandr.

Issue #1085, opened 2025-11-07
[Bug]: Recurring problem with timezones


chrlaney:
### Bug Description

See https://github.com/callumalpass/tasknotes/issues/781 - this was fixed for a while, but has cropped up again. Thanks!

mjkrasny:
Same here with the last ver (4.3.0) of the plugin. All Outlook sourced ICS calendar event are shifted in time.

Update [21/01/2026]: after resetting outlook link, and repeated setup in Obsidian it seems to be working fine now.

callumalpass:
I rechecked this against current main. The core ICS conversion still uses `ICAL.Time.toUnixTime()` for timed events, a direct Outlook `Eastern Standard Time` winter/summer sample converts to the expected UTC instants, and the active #1738 Outlook timezone-alias regression test passes. The only newer follow-up says resetting the Outlook link fixed that user's case, so I could not reproduce a current TaskNotes bug. Thanks again @chrlaney for reporting this and @mjkrasny for the follow-up.

chrlaney:
@callumalpass many thanks for the fix!

Issue #781, opened 2025-09-29
[Bug]: Advanced Calendar doesn't translate between time zones


chrlaney:
### Bug Description

I have subscribed to my Outlook calendar. An event that was scheduled for 3 pm EST by my colleague correctly shows up in my Outlook calendar as 1 pm MST (my time zone), but shows up in Advanced Calendar as 3 pm MST. So it seems that Advanced Calendar is just finding the original time and ignoring time zones.

chrlaney:
Thank you for the speedy fix! Great package.

PR #1913, opened 2026-05-20
fix(ics): resolve Windows TZIDs from Outlook published calendars

kapej42:
Fixes the bugs where ICS subscriptions from Microsoft Outlook (`/owa/calendar/.../calendar.ics`) display events at the wrong time — typically with a fixed offset equal to the user's local zone (e.g. +2h in CEST, +1h in CET).

Refs #781, #1085.

## Repro

Outlook's published-calendar feeds reference events with Windows-style TZIDs and only inline a `VTIMEZONE` block for *some* of those TZIDs — typically the calendar owner's primary zone. Events from invitees in other zones reference TZIDs with no accompanying `VTIMEZONE`. Example from a real feed:

```ics
BEGIN:VTIMEZONE
TZID:Romance Standard Time
...
END:VTIMEZONE

BEGIN:VEVENT
SUMMARY:Tech Lunch
DTSTART;TZID=W. Europe Standard Time:20260520T093000   ← no VTIMEZONE for this!
DTEND;TZID=W. Europe Standard Time:20260520T103000
END:VEVENT
```

The user, viewing in Europe/Amsterdam (CEST = UTC+2), sees this event at **11:30** instead of **09:30**.

## Root cause

`registerCalendarVTimezones` correctly registers `Romance Standard Time` (because the feed inlines its VTIMEZONE) but `W. Europe Standard Time` is unknown to ical.js. `ICAL.Event#startDate.zone` becomes `floating`. `icalTimeToISOString` then calls `toUnixTime()` on a floating time, which interprets the wall clock as UTC. When that ISO string is rendered in the user's local zone, it surfaces as a fixed offset (= user's UTC offset).

ical.js 2.2.1 has no IANA tzdata built in (`ICAL.TimezoneService.has('Europe/Berlin')` is `false` until a VTIMEZONE is registered), so it can't recover from a missing VTIMEZONE on its own.

## Fix

Add a fallback that activates only when ical.js has demoted a time to floating:

1. Read the raw `TZID` parameter from the source property (DTSTART/DTEND/recurring instances).
2. Resolve it via the CLDR `windowsZones` table to an IANA name (e.g. `W. Europe Standard Time` → `Europe/Berlin`), or pass through if it's already a valid IANA name.
3. Convert the wall clock to a UTC instant using `Intl.DateTimeFormat`, which is fully populated with IANA tzdata in Electron/Obsidian.

The existing VTIMEZONE-resolved path remains the fast path — fallback only runs when `zone.tzid === 'floating'`.

### Files

- `src/utils/icsTimezoneFallback.ts` (new) — CLDR map + Intl-based wall-to-UTC conversion
- `src/services/ICSSubscriptionService.ts` — wire the fallback through `icalTimeToISOString`, including the recurring-instance and modified-instance paths
- `src/types/ical.d.ts` — add `Component.getFirstProperty` and a typed `Time.zone` shape
- `tests/__mocks__/ical.ts` — extend the mock with `getFirstProperty` for parity with real ical.js
- `tests/unit/utils/icsTimezoneFallback.test.ts` (new) — 11 unit tests covering the Windows→IANA mapping, IANA pass-through, Intl-based wall-to-UTC conversion across CEST/CET/EST/PDT/UTC/Etc-GMT zones, and DST-boundary behavior

### Trade-offs

- **No new dependencies.** CLDR map is ~130 entries (~3 KB), Intl conversion is ~30 lines.
- **DST ambiguity:** during the spring-forward gap (a wall time that doesn't exist) and fall-back overlap (a wall time that exists twice), the algorithm resolves to one consistent interpretation per iteration. The 3-iteration loop guarantees convergence for normal cases.
- **Behavior preserved for working feeds:** events with proper VTIMEZONE blocks (Google, Cozi, Apple, Thunderbird, …) go through the original `toUnixTime()` path untouched.

## Testing

- ✅ 11/11 new unit tests pass.
- ✅ Full unit suite: 3404 passing, same 4 failures as baseline (pre-existing, unrelated to ICS).
- ✅ Typecheck clean (the 2 unrelated `releaseNotes` errors are pre-existing for the build-time generated file).
- ✅ Verified against a real Outlook-published feed (40 KB, 661 VEVENTs): parses 623 events with 0 errors, and times now match what Outlook itself displays (09:30 events show at 09:30, previously they showed at 11:30 in CEST).

### DIFF
--- a/src/services/ICSSubscriptionService.ts
+++ b/src/services/ICSSubscriptionService.ts
@@ -8,6 +8,7 @@ import type { InterpolationValues, TranslationKey } from "../i18n";
 import { stringifyUnknown } from "../utils/stringUtils";
 import { createTaskNotesLogger } from "../utils/tasknotesLogger";
 import { publishUserNotice } from "../core/userNotices";
+import { resolveTzidToIANA, wallTimeInZoneToUtcIso } from "../utils/icsTimezoneFallback";
 
 const tasknotesLogger = createTaskNotesLogger({ tag: "Services/ICSSubscriptionService" });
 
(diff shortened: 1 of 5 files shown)


callumalpass:
Thanks @kapej42 . Merging for next release. Much appreciated.

PR #1969, opened 2026-05-29
feat: add setting to hide subtasks from task list root

### LINKED ISSUE #1715: [FR]: Hide subtasks when they're folded under Main Task
### Feature Description

Currently, I have enabled the feature to show subtasks folded under the main task in a list view, which is very helpful.  However, they also show up separately in the list as independent tasks.  I would prefer either as default or option to not have them show up in both places.

Spencerduran:
## Summary

Closes #1715

Adds a **"Hide subtasks in list root"** toggle in Settings → Appearance (alongside the existing expandable subtasks options). When enabled, tasks that belong to a parent project are hidden from the root level of task list views — they remain visible only when their parent task is expanded.

## Implementation

Pure CSS, no JS-level deduplication required. The feature works by toggling a `tasknotes-hide-root-subtasks` class on `document.body`, which activates this rule in `task-list-view.css`:

\`\`\`css
body.tasknotes-hide-root-subtasks .tasknotes-plugin
  .task-card:has(.task-card__project-link):not(:has(.task-card__subtasks)):not(.task-card__subtasks .task-card) {
    display: none;
}
\`\`\`

The three-part selector:
- `:has(.task-card__project-link)` — matches tasks that belong to a parent
- `:not(:has(.task-card__subtasks))` — excludes expanded parent tasks (which have a subtasks container), preventing them from being hidden
- `:not(.task-card__subtasks .task-card)` — excludes subtasks already nested under an expanded parent, so they still show in the fold-out

## Changes

- `src/types/settings.ts` — adds `hideRootSubtasks: boolean`
- `src/settings/defaults.ts` — defaults to `false` (opt-in)
- `src/settings/tabs/appearanceTab.ts` — adds toggle with body class side-effect
- `src/main.ts` — applies/removes body class on load/unload
- `src/i18n/resources/en.ts` — adds English strings
- `styles/task-list-view.css` — adds the CSS rule

## Test plan

- [ ] Enable the setting — subtasks at root level disappear
- [ ] Expand a parent task — subtasks appear in the fold-out as expected
- [ ] Collapse the parent — subtasks return to hidden
- [ ] Disable the setting — root-level subtasks reappear
- [ ] Reload Obsidian with setting enabled — body class applied on startup, subtasks hidden immediately

### DIFF
--- a/src/i18n/resources/en.ts
+++ b/src/i18n/resources/en.ts
@@ -1566,6 +1566,11 @@ export const en: TranslationTree = {
 					name: "Show expandable subtasks",
 					description: "Allow expanding/collapsing subtask sections in task cards",
 				},
+				hideRootSubtasks: {
+					name: "Hide subtasks in list root",
+					description:
+						"In task list views, hide tasks that belong to a parent project from the root level. Subtasks remain visible when their parent is expanded.",
+				},
 				expandSubtasksByDefault: {
 					name: "Expand subtasks by default",
 					description: "Show project subtasks expanded when task cards are rendered",
--- a/src/main.ts
+++ b/src/main.ts
@@ -261,6 +261,7 @@ export default class TaskNotesPlugin extends Plugin {
 		});
 
 		await this.loadSettings();
+		document.body.classList.toggle("tasknotes-hide-root-subtasks", this.settings.hideRootSubtasks);
 		this.performanceProfiler = createTaskNotesPerformanceProfiler({
 			isEnabled: () => this.settings?.enableDebugLogging === true,
 			logger: createTaskNotesLogger({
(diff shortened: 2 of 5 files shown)


Spencerduran:
Just noticed #1792 tackles the same problem with a JS-based approach. It filters subtasks before rendering by checking whether each task's parent is in the current result set, so subtasks stay visible when their parent is filtered out. This PR is a global CSS toggle with no per-view config and no awareness of whether the parent is in the view.

Happy to close this if you'd prefer #1792.

callumalpass:
Thanks @Spencerduran for submitting this PR. I've decided to implement a separate approach that is now available in 4.10.0. Very much appreciate the PR here, though!

PR #2019, opened 2026-06-09
Fix: Kanban "Hide empty columns" ignored when swimlanes are enabled

### LINKED ISSUE #2018: [Bug]: Kanban view - when swimlanes are enabled, "Hide empty columns" doesnt work
### Bug Description

Kanban view "Hide empty columns" doesnt work when swimlanes are enabled

Expected: empty columns disappear
Actual: empty columns remain

Quick analysis: looks like `buildKanbanSwimlaneColumns` doesnt call `shouldRenderKanbanColumn` like `renderFlat` does?

### Steps to Reproduce

1. make a kanban view, 
2. enable swimlanes, 
3. group-by another field for columns, and 
4. enable "Hide empty columns".  
5. Then apply a filter to empty out a column (0 tasks)

### TaskNotes Debug Info

```text

```

### Affected Task Frontmatter

```yaml

```

```yaml

```

_No response_

### Console Errors or Logs

```text

```

Image

scottTomaszewski:
## Summary

"Hide empty columns" had no effect in the Kanban view when swimlanes were enabled. Flat (no-swimlane) mode filters each column through `shouldRenderKanbanColumn` in `renderFlat`, but the swimlane path (`renderWithSwimLanes` → `renderSwimLaneTable`) rendered every column key unconditionally, so emptied columns stayed visible in both the header row and every swimlane row.

- Add `getVisibleKanbanSwimLaneColumnKeys` to `kanbanGrouping.ts`, which uses the existing `getKanbanColumnTaskCounts` to total each column across all swimlanes and drops columns that are empty (and not pinned) when "Hide empty columns" is on.
- Wire it into `renderWithSwimLanes`, passing the filtered column keys to `renderSwimLaneTable`. Counts are taken from the post-filter `swimLanes` map so they reflect the active filter; pinned columns are preserved, matching flat-mode behavior.

## Test Plan

- [x] Unit tests for the new helper in `tests/unit/bases/kanbanGrouping.test.ts` (hides empty columns, keeps empty pinned columns, keeps all columns when the option is off).
- [x] Regression test `tests/unit/issues/issue-kanban-hide-empty-columns-swimlanes.test.ts` covering the build-groups → build-swimlanes → filter pipeline.
- [x] `tsc --noEmit` passes.
- [x] Manual: create a Kanban view, enable swimlanes, group columns by one field and swimlanes by another, enable "Hide empty columns", then filter a column to 0 tasks → the column disappears from headers and all swimlane rows while pinned columns remain.

## Reproduction

1. Make a Kanban view.
2. Enable swimlanes; group columns by another field.
3. Enable "Hide empty columns".
4. Apply a filter that empties a column (0 tasks). Before this change the empty column remained; now it is hidden.

Resolves #2018


scottTomaszewski:
Works: 

image

Issue #2043, opened 2026-06-18
Default Subtasks "+" button does not assign current project (generated filter not parsed by filter-defaults)


kudrmax:
### Bug Description

The "+ New" button in the **default Subtasks tab** of the Relationships widget (and in the generated `relationships.base`) creates a task that is **not** assigned to the current project, so the task immediately disappears from the view. This is the same symptom as #1657 (fixed in 4.8.0 for simple user-authored filters), but it still happens for the filter that **TaskNotes itself generates by default** — so the 4.8.0 fix does not cover the out-of-the-box subtasks view.

Reproduced on 4.11.0.

### Root cause

When the "+" button is used, `KanbanView.openTaskCreationForKanbanCell` → `createFileForView` → `extractBasesFilterDefaults` derives the new task's frontmatter **only** from the view filter. The parser in `src/bases/basesFilterDefaults.ts` recognizes these deterministic forms:

- `file.hasTag("x")`
- `prop == "x"`
- `prop.contains("x")`
- `prop.contains(this.file.asLink())`

But the default relationship "Subtasks" filter generated in `src/templates/defaultBasesFiles.ts` (around line 959, via `formatProjectEntryLinkExpression`) looks like:

```
file.hasLink(this.file) && list(note.projects).map(file(value.replace(/^\[[^\]]+\]\((.*)\)$/, "$1").replace(/%20/g, " ")).asLink()).contains(this.file.asLink())
```

The `currentFileContainsMatch` regex `^(.+?)\.contains\(this\.file\.asLink\(\)\)$` does match the whole string, but `normalizeFilterProperty` then receives the prefix `file.hasLink(this.file) && list(note.projects).map(...)` and cannot reduce it to the field name `projects`, so it returns `null` and no `projects` default is applied.

Net effect: the created task has no `projects`, fails the view filter, and vanishes from the widget. The button looks broken.

So the plugin generates a normalized filter (to tolerate markdown-style links / `%20` / paths) that its own filter-defaults parser cannot read.

### Steps to reproduce

1. Use default settings (tag-based `#task`, default field mapping).
2. Open any note used as a project (i.e. a task links to it via `projects`).
3. In the Relationships widget, open the default **Subtasks** (Kanban) tab.
4. Click the column "+" button, name the task, save.
5. The task is created without a `projects` value and disappears from the Subtasks view.

### Workaround

Simplify the Subtasks tab filter in `relationships.base` to a form the parser understands:

```yaml
filters:
  and:
[...]

Issue #1657, opened 2026-03-01
[Bug]: "+ New" button on Kanban view doesn't assign new task to current project


casualQuads122:
### Bug Description

I have a project page and a TaskNotes Bases Kanban view for all my tasks in that project.
When I click the "+ New" button on the upper right-hand corner, the new task modal from TaskNotes pops up.
After naming the task and saving it, it isn't assigned to the project I'm in, it just doesn't have a projects property.

<img width="728" height="43" alt="Image" src="https://github.com/user-attachments/assets/ed6bdc48-031b-43b9-a6dc-fec79aa4dc95" />

I have to rely on using checkboxes to add tasks to the project.

callumalpass:
Thank you for reporting this. This should be resolved in release 4.8.0, so I am going to close this issue now.

PR #2044, opened 2026-06-18
fix(bases): parse generated normalized relationship filter for task defaults (#2043)

kudrmax:
## Summary

Fixes the default Subtasks "+" button not assigning the current project (see #2043). This is a follow-up to #1657: the filter-defaults parser added in 4.8.0 handles simple user-authored filters, but not the more complex filter that TaskNotes generates by default for relationship views.

## Problem

Creating a task via the column "+" button derives the new task's frontmatter from the view filter through `extractBasesFilterDefaults`. The default Subtasks relationship view (generated in `src/templates/defaultBasesFiles.ts`) normalizes each project link before comparing, producing:

```
file.hasLink(this.file) && list(note.projects).map(file(value.replace(/^\[[^\]]+\]\((.*)\)$/, "$1").replace(/%20/g, " ")).asLink()).contains(this.file.asLink())
```

`normalizeFilterProperty` cannot reduce that prefix to the field name `projects`, so it returns `null`, no `projects` default is applied, and the created task immediately drops out of the view.

## Change

`applyFilterRuleDefault` now detects the normalized `list(note.PROP).map(...).contains(this.file.asLink())` form (with an optional `... && ` prefix) and recovers the wrapped property via the existing `normalizeFilterProperty` (which already unwraps `list(...)` and `note.`/`task.`). The current note link is then pre-filled as the project, so the new task stays in the view.

The change is targeted: it only triggers on the `list(...).map(...).contains(this.file.asLink())` shape and falls through to the existing generic matcher otherwise. Field mapping is respected (the property is resolved through `fieldMapper`).

## Tests

Added two regression tests in `tests/unit/bases/basesFilterDefaults.test.ts`:
- extracts `projects` from the exact generated Subtasks filter;
- honors a custom `projects` field mapping for the same shape.

`npx jest tests/unit/bases/` passes (39 suites, 232 tests). Pre-existing typecheck errors about a build-generated `../releaseNotes` module are unrelated to this change.

### DIFF
--- a/src/bases/basesFilterDefaults.ts
+++ b/src/bases/basesFilterDefaults.ts
@@ -116,6 +116,23 @@ function applyFilterRuleDefault(
 		return;
 	}
 
+	// Generated relationship views (e.g. the default Subtasks tab) normalize each
+	// link before comparing, producing a rule such as:
+	//   file.hasLink(this.file) && list(note.projects).map(...).contains(this.file.asLink())
+	// The generic matcher below cannot reduce that prefix to a field name, so detect
+	// the normalized list/map form explicitly and recover the wrapped property.
+	const mappedListContainsCurrentFileMatch = trimmedRule.match(
+		/(?:^|&&\s*)(list\((?:note\.|task\.)?[\w-]+\))\.map\([\s\S]*\)\.contains\(this\.file\.asLink\(\)\)$/
+	);
+	if (mappedListContainsCurrentFileMatch) {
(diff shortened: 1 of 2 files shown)


callumalpass:
Thanks for contributing this, @kudrmax. The Subtasks-defaults issue is now fixed in 4.13.0 via #2055 and #2323, so I'm closing this PR as superseded. Much appreciated!

Issue #2046, opened 2026-06-18
[Bug]: Quick Actions for Task Under Cursor "Set due date" autofocuses incorrectly


DevOps-Toast:
### Bug Description

When trying to use only keyboard hotkeys to interact with TaskNotes, the command "Quick actions for task under cursor" is very useful. A small annoyance is NLP box is not focused on pop-up, the regular date selector is.

### Steps to Reproduce

1. Create any new task
2. Link to that task in a note
3. With cursor inside link, run "Quick actions for task under cursor"
4. Type "Set due date" or "Set scheduled date", hit enter
5. Observe the pop-up skipping over the NLP box and auto-focusing the date box

<img width="414" height="225" alt="Image" src="https://github.com/user-attachments/assets/3850ea7d-bc2e-40bf-983f-ac357588c35c" />

### TaskNotes Debug Info

```text

```

### Affected Task Frontmatter

```yaml

```

### Relevant Bases File Content

```yaml

```

### Console Errors or Logs

```text

```

Issue #2045, opened 2026-06-18
[Bug]: Custom boolean field doesn't show in Basic Information


chmac:
### Bug Description

When I try to add a custom field of type boolean to the Basic Information section, the field does not appear in the UI to edit a task.

### Steps to Reproduce

1. add a custom field of type boolean
2. open settings > modal fields
3. click sync user fields
4. move custom field into the basic information section

### TaskNotes Debug Info

```text

```

### Affected Task Frontmatter

```yaml

```

### Relevant Bases File Content

```yaml

```

### Console Errors or Logs

```text

```

### Screenshots or Screen Recording

<img width="1120" height="841" alt="Image" src="https://github.com/user-attachments/assets/eb0a8f53-1975-4439-b896-0f6c145a6ca5" />
<img width="810" height="913" alt="Image" src="https://github.com/user-attachments/assets/ed93a0b2-5d81-4b9b-a03f-fde5b0ca4973" />

callumalpass:
Resolved in 4.11.1. Thanks @chmac for reporting this; I’m closing this issue.

PR #2052, opened 2026-06-21
fix(modal-fields): render user fields assigned to the basic group (issue #2045)

ther12k:
## Summary

When a user assigns a custom (user-typed) field to the 'Basic Information' group via the Modal Fields settings, the field silently disappears. The cause was a hardcoded skip in `renderTaskModalFieldGroups`:

```ts
if (groupConfig.id === "basic") {
    continue;
}
```

The skip was originally added to avoid double-rendering the built-in basic fields (title, details), which have dedicated creation paths in `TaskModal` (`createTitleInput`, `detailsMarkdownEditor`). Unfortunately, the same skip also discarded any user field the user moved into the basic group.

This change:
1. Removes the hardcoded basic-group skip.
2. Relies on the existing `renderTaskModalField` dispatcher to silently drop core basic fields (they are not in `fieldRenderers`, and they are not user fields, so `renderTaskModalField` returns `false` and they get counted in `ignoredFieldIds`). Their UI surfaces remain unchanged because the dedicated creation paths still run elsewhere.
3. Lets user fields in the basic group render through `renderUserField` exactly like they do in every other group.
4. Adds a guard at the end of the group loop that removes an empty `task-modal__field-group` container when the group produced zero rendered fields. This keeps the DOM clean for the basic group when it contains only core fields, and is also defensive against any future group whose fields are all core-only.

## Tests

Updates `tests/unit/modals/taskModalFieldRenderer.test.ts` to reflect the new (correct) behavior — the previous expectations encoded the bug. Both tests now correctly report `title`/`details` as ignored, and assert that empty group containers are removed.

Adds `tests/unit/issues/issue-2045-user-fields-in-basic-group.test.ts` with three cases:
- User field moved into the basic group renders through `renderUserField`
- Groups whose fields all render keep their container (no regression)
- Core-only basic group produces no leftover empty container

All new + updated tests pass. Full `tests/unit/modals` suite (37 suites, 187 tests) passes. `tsc --noEmit --skipLibCheck` clean. `npm run lint` clean.

## Verification

```bash
npm install
node generate-release-notes-import.mjs
npx tsc -noEmit --skipLibCheck
npx jest tests/unit/modals --no-coverage
node scripts/run-lint.mjs
```

## Note

This PR is separate from PR #2051 (issue #2046 NLP input autofocus). They were developed in parallel as independent fixes.

### DIFF
--- a/src/modals/taskModalFieldRenderer.ts
+++ b/src/modals/taskModalFieldRenderer.ts
@@ -60,12 +60,9 @@ export function renderTaskModalFieldGroups(
 	};
 
 	for (const groupConfig of groupsToRender) {
-		if (groupConfig.id === "basic") {
-			continue;
-		}
-
 		const groupContainer = options.container.createDiv({ cls: "task-modal__field-group" });
 		result.groupsRendered += 1;
+		let fieldsRenderedInGroup = 0;
 
 		for (const field of groupConfig.fields) {
 			const rendered = renderTaskModalField({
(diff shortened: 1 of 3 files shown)


callumalpass:
Thanks for putting this together and for the thorough regression coverage. The same fix was independently implemented in `846d328d` and has since shipped in TaskNotes 4.11.1, so I’m closing this PR as superseded. Your diagnosis and proposed approach were sound—appreciate the contribution!

PR #2053, opened 2026-06-21
fix: do not enqueue Google Calendar sync when export is disabled (closes #2040)

### LINKED ISSUE #2040: [Bug]: googleCalendarSyncQueue fills up without a calendar configured
### Bug Description

I just installed the plugin, created some tasks, and I see a growing `googleCalendarSyncQueue` property in `data.json`. I realised because I track my vault with git, now every time I create a new task it seems like it adds an entry into that array. But I don't have any plans to sync to google calendar.

### Steps to Reproduce

1. install plugin
2. git commit `data.json`
3. create 3 tasks
4. note that `data.json` has changed

### TaskNotes Debug Info

```text
Every line shows:

      "lastError": "Google Calendar sync is not ready"
```

### Affected Task Frontmatter

```yaml

```

```yaml

```

_No response_

### Console Errors or Logs

```text

```

I verified by opening the sandbox vault, and I see the same behaviour.

ther12k:
Fixes #2040.

**Root cause**
`syncTaskToCalendar` in `src/services/TaskCalendarSyncService.ts` enqueues entries into `googleCalendarSyncQueue` even when `settings.googleCalendarExport.enabled === false`. The check happens later in the pipeline, after the queue interaction, so disabled users see queue churn and retry attempts they never asked for.

**Fix**
Early-return at the top of `syncTaskToCalendar` when `settings.googleCalendarExport.enabled === false`, before any queue interaction. Transient-failure queueing for users who *do* have export enabled is preserved.

**Tests**
- Added 3 new tests in `tests/unit/issues/issue-2040-google-calendar-sync-queue.test.ts`:
  1. disabled user → no enqueue, no retry attempt
  2. enabled user → still enqueues on transient failure (regression guard)
  3. enabled user → completes normally without queue growth
- All 38 sync-service suites pass.

**Notes**
- Scoped to #2040 only. Independent from #2051 (#2046, NLP autofocus) and #2052 (#2045, Basic Information user fields).

### DIFF
--- a/src/services/TaskCalendarSyncService.ts
+++ b/src/services/TaskCalendarSyncService.ts
@@ -2425,6 +2425,16 @@ export class TaskCalendarSyncService {
 		}
 
 		const settings = this.plugin.settings.googleCalendarExport;
+		// If the user has not enabled Google Calendar export at all, do nothing
+		// and do not queue. The queue is for transient failures that might be
+		// recoverable (auth expired, rate limited, target calendar temporarily
+		// missing). When the user has opted out of the feature entirely, every
+		// eligible task creation would otherwise pollute the queue with an
+		// entry that will never be processed (issue #2040).
+		if (!settings.enabled) {
+			return true;
+		}
+
 		const existingEventId = this.getTaskEventId(task);
 		const targetCalendarId = settings.targetCalendarId;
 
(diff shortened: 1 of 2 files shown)


callumalpass:
Thanks for investigating this and for adding the regression tests. This issue was addressed in #2041, which is now merged into `main`, so I’m closing this PR as superseded.

Thanks again for the contribution!

Issue #1902, opened 2026-05-19
[Bug]: Markdown-style links don't snow as subtasks on Project pages


minchinweb:
### Bug Description

If a task file has a wikilink (e.g.

```
projects:
  - "[[3D Printing]]"
```
), it will show as a subtask on the project page. But if it is a Markdown (classical) style link, it doesn't show:

```
projects:
  - "[3D Printing](../../../projects/3D%20Printing.md)"
```

My "SubTasks" view is slightly modified, but I think the only thing I did was change it from a Kanban Board to a TaskList. This is the view is below. But these are also missing from the default (Kanban Board) SubTasks view.

**So I think `list(note.projects).contains(this.file.asLink())` is failing on Markdown links.**

### Steps to Reproduce

Create a note to use as a project.

Create a tasknote that uses a Markdown link to point to that project.

Notice the missing task on the project page.

### TaskNotes Debug Info

```text
SYSTEM INFO:
	Obsidian version: 1.12.7
	Installer version: 1.12.7
	Operating system: Windows 10 Pro 10.0.19045
	Login status: not logged in
	Language: en
	Insider build toggle: off
	Live preview: on
	Base theme: adapt to system
	Community theme: Minimal 8.2.0
	Snippets enabled: 29
	Restricted mode: off
	Plugins installed: 48
	Plugins enabled: 27
		1: Calendar v1.5.10
		2: Tab Switcher v1.4.0
		3: Linter v1.31.2
		4: Note Refactor v1.8.2
		5: Open vault in VS Code v1.4.2
		6: Periodic Notes v0.0.17
		7: Tag Wrangler v0.6.4
		8: Better Word Count v0.10.1
		9: Plugin Update Tracker v1.7.0
		10: Style Settings v1.0.9
		11: Minimal Theme Settings v8.2.3
		12: Advanced Tables v0.23.2
		13: Sort & Permute lines v0.7.0
		14: Pane Relief v0.5.9
		15: Show Current File Path v0.5.2
		16: TaskNotes v4.8.0
		17: Omnisearch v1.29.2
		18: Extended Markdown Syntax v2.0.11
		19: Quick Switcher++ v6.1.2
		20: Map View v6.1.4
		21: Footnote Shortcut v0.1.3
		22: Colored Tags v6.1.2
		23: Better Properties v1.0.0
		24: Timekeep v2.1.3
		25: Tracker v1.19.0
		26: Git v2.38.3
[...]

PR #2055, opened 2026-06-21
fix(bases-filter-defaults): extract property from generated conjunction + map filter (closes #2043)

### LINKED ISSUE #2043: Default Subtasks "+" button does not assign current project (generated filter not parsed by filter-defaults)

ther12k:
Fixes #2043.

**Root cause**
The default-relationships Subtasks filter generated by `formatProjectEntryLinkExpression` in `src/templates/defaultBasesFiles.ts` emits a normalized expression like:

```
file.hasLink(this.file) && list(note.projects).map(file(value.replace(/^\[[^\]]+\]\((.*)\)$/, "$1").replace(/%20/g, " ")).asLink()).contains(this.file.asLink())
```

`currentFileContainsMatch` in `src/bases/basesFilterDefaults.ts` matches the entire conjunction as the property expression. `normalizeFilterProperty` then sees:

```
file.hasLink(this.file) && list(note.projects).map(file(value.replace(/^\[[^\]]+\]\((.*)\)$/, "$1").replace(/%20/g, " ")).asLink())
```

The leading `list(` regex doesn't match (string starts with `file.hasLink`), the generated `.map(...)` wrapper prevents recognition, and `note.`/`task.` prefix stripping leaves a complex expression that doesn't match any core/user field — so `normalizeFilterProperty` returns `null` and no `projects` default is applied. Net effect: the task is created without a `projects` field, fails the view filter, and vanishes from the Subtasks widget.

**Fix**
In `normalizeFilterProperty`, before the existing `list(...)` / `note.` / `task.` recognition runs:

1. If the expression contains `&&`, keep only the right-hand operand (strips the `file.hasLink(this.file) &&` left side).
2. If the expression contains a trailing `.map(...)`, strip it (balanced-paren walk, since `.asLink()` lives inside the `.map` argument, not after it).

After these two steps, the existing logic cleanly extracts `projects` (or any other core/user field) from the normalized list/map expression.

**Tests**
- 3 new regression tests in `tests/unit/bases/basesFilterDefaults.test.ts`:
  1. core `projects` field extracted from the generated conjunction + map filter
  2. user-defined `customField` extracted from the generated conjunction + map filter
  3. graceful fallback (empty defaults) when no current file link is available
- All 6/6 tests in `basesFilterDefaults.test.ts` pass.
- All 13/13 tests across `kanbanCreationDefaults`, `basesTaskCreation`, `KanbanView.manualOrderFastPath`, and `basesCreateFileForView` pass — no regressions in adjacent code paths.

**Scope**
Scoped to #2043 only. Independent from #2051 (#2046, NLP autofocus), #2052 (#2045, Basic Information user fields), and #2053 (#2040, Google Calendar sync queue).

Refs #2043, #1657, #1902.


callumalpass:
Thanks for contributing this fix. It’s now merged into main with some follow-up adjustments. Much appreciated!

PR #2102, opened 2026-07-05
Fix Google Calendar Obsidian link escaping

martin-forge:
## Summary
- Escape the Obsidian URI query separator in Google Calendar sync descriptions as `&amp;file=`.
- Keep the link plain text and update the focused service test plus release note.

## Why
Google Calendar descriptions are HTML-capable. Leaving the raw `&file=` separator in the description can be handled as an HTML entity boundary by Calendar clients, which breaks the Obsidian open link path parameter.

## Testing
- `npm test -- --runTestsByPath tests/services/TaskCalendarSyncService.test.ts`
- `npm run build:test`
- `npm run lint`
- Upstream CI passed: docs, test (20), and build-test.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -31,3 +31,8 @@ When a change has user-facing documentation, include a canonical tasknotes.dev l
 ```
 
 -->
+
+## Fixed
+
+- Fixed Obsidian links in Google Calendar event descriptions so the note path survives HTML entity handling.
+  - Thanks to @martin-forge for reporting and fixing.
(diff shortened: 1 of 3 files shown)


martin-forge:
Closing this — I tested the premise and it does not hold.

I created two Google Calendar events, one storing a raw `&` in the description and one storing `&amp;`, and read both back through the API. Google stores the field byte-for-byte and normalises nothing. Rendering then differs by client:

- Apple Calendar on macOS decodes `&amp;` to `&` and auto-linkifies the URL, so both events displayed an identical working link. The raw `&` was never broken there.
- TaskNotes' own calendar event modal renders the description as plain text and does not decode entities, so `&amp;` appears literally and the Obsidian URI is broken.

So this change fixes nothing I could reproduce and breaks the link in TaskNotes itself. Apologies for the noise.

PR #2187, opened 2026-08-02
perf: don't block onload on Bases registration

tgrosinger:
registerBasesTaskList retried registration five times with 200ms sleeps when the Bases plugin had not finished loading, and onload awaited the whole thing. On mobile, where plugin load ordering is less predictable, that could stall plugin startup for a full second.

registerBasesIntegration already owns retrying: it schedules a timer that re-attempts until the views are actually registered, checks the Bases plugin is enabled before each try, and is torn down on unload. So drop the inner loop entirely and make registerBasesTaskList a single attempt that reports its result, leaving one retry mechanism instead of two. onload no longer awaits it either.


callumalpass:
Thanks for this @tgrosinger! Merging for next release. Much appreciated

PR #2191, opened 2026-08-03
Fix moved recurring occurrences rescheduling their original date

martin-forge:
## Problem

A moved occurrence replaces its original slot in the recurrence, but next-occurrence planning ignored that stored exclusion. Completing or skipping an occurrence that was moved *earlier* than its original date could therefore schedule the original date again.

| Series | Move | Complete or skip | Before | After |
| --- | --- | --- | --- | --- |
| Weekly | 13 April → 10 April | 10 April | 13 April | 20 April |
| Daily | 13 April → 10 April | 10 April | 11, 12, **13**, 14 April | 11, 12, 14 April |

In the daily case the replaced date reappeared after intervening occurrences, so the series silently gained an extra instance.

## Fix

Treat pending and resolved moved-original dates as already processed while calculating the next open occurrence. This applies only to that calculation — persisted completion and skip history stays keyed to the moved date, so no stored task state changes shape and nothing needs migrating.

Completion-anchored recurrence is untouched; the adjustment is scoped to `scheduled`-anchored series, where the moved-original date is what the planner searches against.

## Validation

```
npm test -- tests/unit/services/taskRecurringPlanning.test.ts
npm run typecheck
npm run lint
npm run build
```

16 focused tests pass, including cases covering weekly and daily series, later moves (unchanged behaviour), and completion-anchored recurrence. Rebased on current `main`.


callumalpass:

PR #2192, opened 2026-08-03
Fix provider calendar source names on event cards

martin-forge:
## Problem

Agenda and list cards already use their metadata line for the source calendar: ICS events show the subscription name, while the calendar icon separately identifies the row as a calendar event.

Google and Microsoft events missed that lookup and fell back to the generic "Calendar" label, even when TaskNotes already knows the calendar's name. With several provider calendars visible at once, every one of those rows reads "Calendar".

## Fix

Resolve provider-owned events through `CalendarProviderRegistry`, so the existing source field can show names such as Personal or Work. Google's `primary` alias is mapped to the primary calendar's metadata, since events on the default calendar report that alias rather than the calendar's own id.

Unchanged: the calendar-event icon label, ICS subscription names, and the translated fallback for events that neither a subscription nor a provider owns.

Both card paths already look the subscription up for the event colour, so that result is passed through rather than repeated — no extra lookup is added to card render or update.

## Validation

```
npm test -- tests/unit/issues/issue-provider-calendar-source-name.test.ts
npm run typecheck
npm run lint
npm run build
```

Rebased on current `main`.


callumalpass:

PR #2193, opened 2026-08-03
Fix raw HTML in Google Calendar event descriptions

martin-forge:
## Problem

The Google Calendar API documents the event `description` field as one that can contain HTML, and Calendar's own editor stores markup there. TaskNotes consumes that string as plain text everywhere: event tooltips, the ICS event info modal, "copy as markdown", the `{{icsEventDescription}}` template variable, generated note bodies, and folder templates.

Several of those write the value into the vault, so raw markup was not only displayed — it was persisted into note bodies and folder names.

## Fix

Normalize the description once at the Google provider boundary, parsing it with Obsidian's HTML sanitizer. Paragraph breaks, list structure, and link addresses survive the flattening.

Plain text is returned unchanged, so descriptions that merely contain angle brackets — `Contact <user@example.com>`, placeholders such as `<TBC>` — are untouched. Normalizing at the boundary keeps every consumer plain-text rather than making each one handle HTML separately. ICS subscriptions read descriptions through their own path and are unaffected.

### Why not a plain `textContent` read

`textContent` drops paragraph and list breaks and discards `href` values, so a description holding a bulleted agenda and a meeting link collapses into one run-on line with the link gone. The flattening pass exists to keep those three things and nothing more.

### Descriptions sent back to Google

Writes are unaffected. `TaskCalendarSyncService` rebuilds the description from the task before every update, and the drag/resize path sends only start and end. The normalized value is never round-tripped into Calendar, so formatting a user has set there is not flattened by this change.

## Relationship to earlier work

This complements #1882, which fixed the TaskNotes-to-Google export; this covers the Google-to-TaskNotes input side. It is also the actual cause behind #2102, which I closed after testing: the entity handling I proposed there was a symptom of descriptions being read as plain text, not the underlying problem.

## Validation

```
npm test -- tests/unit/utils/calendarDescription.test.ts tests/services/GoogleCalendarService.test.ts
npm run typecheck
npm run lint
npm run build
```

50 focused tests pass. The full suite shows no new failures against current `main`. Rebased on current `main`.

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -35,3 +35,7 @@ When a change has user-facing documentation, include a canonical tasknotes.dev l
 ## Fixed
 
 - Fixed `PUT /api/tasks/:id` ignoring empty arrays for `contexts` and `blockedBy`: sending `{"contexts": []}` or `{"blockedBy": []}` now clears the corresponding frontmatter field instead of silently leaving the previous value in place. The deletion pass previously fired only on a literal `undefined`, which JSON cannot express, so HTTP clients had no way to clear these fields. Thanks to @tgrosinger for the contribution.
+- (#2193) Google Calendar event descriptions written with formatting now read as
+  plain text in event details, copied text, and generated notes, instead of
+  showing raw HTML tags. Paragraph breaks, list structure, and link addresses are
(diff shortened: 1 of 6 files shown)


callumalpass:

PR #2221, opened 2026-08-11
Clear contexts and blockedBy on empty-array updates

tgrosinger:
The frontmatter deletion pass in removeUnsetMappedFields only removed the contexts and blockedBy keys when the update held a literal undefined, which JSON cannot express. HTTP clients sending `PUT {"contexts": []}` or `PUT {"blockedBy": []}` therefore got a 200 with the previous value silently left in place. Use the same `hasOwnProperty` plus empty-array deletion pattern that projects and reminders already use.


tgrosinger:
Thank you!

PR #2291, opened 2026-09-05
fix: re-read frontmatter before modal save to prevent overwriting concurrent task edits

martin-forge:
### Problem
A background update to a task file can be undone when an already-open task edit window saves an unrelated change. For example, changing an estimate can restore an old scheduled date after a recurring task was completed elsewhere.

### Solution
Read the latest frontmatter inside Obsidian's save callback and apply only the requested fields. Preserve background edits, completed occurrences and fields another writer removed. This changes the task update service only; it does not depend on the Calendar or security proposals.

### Testing
- 110 focused tests pass, including a stale-edit regression and existing task/recurrence tests.
- `npm run lint` and `npm run build:test` pass.
- Reloaded the standalone build in a disposable Obsidian vault; no captured errors.
- [GitHub CI: full test suite and build passed](https://github.com/callumalpass/tasknotes/actions/runs/33980772916).

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -34,6 +34,9 @@ When a change has user-facing documentation, include a canonical tasknotes.dev l
 
 ## Fixed
 
+- Preserve background task updates when saving an older task edit window, including recurring completion history and fields removed by another writer.
+  - Thanks to @martin-forge for the contribution.
+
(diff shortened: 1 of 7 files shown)


callumalpass:
Thanks for all these fixes and the careful testing. They’re merged into main now—really appreciate your work.

martin-forge:
You're very welcome, and thank you for merging them and for all your work on TaskNotes.

PR #2293, opened 2026-09-05
feat(calendar): opt-in durable UUID recovery for private Google Calendar projections

martin-forge:
### Problem
Google task events can go missing, duplicate, or link to an old path after task files move or the local event index is lost. A damaged or missing task file must not make recovery delete its Calendar event.

### Solution
Add an optional **Rebuild private task projections** setting, off by default. Exported tasks receive a persistent UUID. Startup and the existing recovery loop use it to repair event links, missing events and stale details.

Cleanup requires a readable task with the same UUID and either a retirement marker or an identical surviving event. Missing IDs and unreadable files preserve events. The same check protects queued retries, and deletion refuses a provider revision that changed after verification. Unrelated and invited events remain untouched.

### Testing
- 30 recovery and provider-safety tests pass, including damaged files, retries, retirement, differing duplicate content and changed provider revisions.
- Lint and `build:test` pass. Live Obsidian checks against disposable Google events confirm that malformed frontmatter and missing task IDs preserve the events.
- Two local filtering/locale failures also reproduce on the unchanged parent branch.
- [GitHub CI passed: tests, lint, type checks and build](https://github.com/callumalpass/tasknotes/actions/runs/33996089980).

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -40,7 +40,6 @@ When a change has user-facing documentation, include a canonical tasknotes.dev l
 ## Added
 
 - Add an optional **Rebuild private task projections** setting to recover missing Google Calendar task events and repair renamed links, duplicates and stale event details. It uses a persistent task ID, leaves unrelated or invited events alone, and skips unchanged events.
-  - Thanks to @martin-forge for the contribution.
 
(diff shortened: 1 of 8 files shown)


martin-forge:
Closing this optional recovery feature for now. It adds a new setting and a substantial reconciliation and cleanup lifecycle, so I would prefer to establish maintainer interest in that design before asking for its review. The branch is retained in case the feature is wanted later.

PR #2295, opened 2026-09-05
fix: keep saved Pomodoro sessions paused on mobile startup

martin-forge:
### Problem
Opening the mobile app can automatically resume a saved Pomodoro session synced from another device, allowing an old session to run without a new start action.

### Solution
Restore the saved session in a paused state on mobile. Keep the session available for the user to resume; desktop startup behavior is unchanged.

### Testing
- Seven focused Pomodoro tests pass, including the mobile startup regression.
- `npm run lint` and `npm run build:test` pass.
- Reloaded the standalone build in a disposable Obsidian vault; no captured errors.
- [GitHub CI: full test suite and build passed](https://github.com/callumalpass/tasknotes/actions/runs/33979266712).

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -34,6 +34,9 @@ When a change has user-facing documentation, include a canonical tasknotes.dev l
 
 ## Fixed
 
+- Keep a saved Pomodoro session paused when opening the mobile app, so synced timer state does not automatically restart an old session.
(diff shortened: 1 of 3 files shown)


martin-forge:
Withdrawing this proposal. Pausing every saved mobile session changes existing timer behavior; I haven't established a bug that justifies that default change. The task-edit, Calendar, OAuth and settings fixes remain independent.

PR #2300, opened 2026-09-08
fix(reading-mode): nest the task card inside the header so virtualisation cannot delete it

### LINKED ISSUE #2255: [Bug]: Reading mode - the task card is deleted and re-injected on every render pass, making the note jump while scrolling
### Bug Description

In reading mode, a note carrying the in-note task card jumps under the scroll, repeatedly and jarringly enough to make reading mode unusable. Only notes carrying the card are affected; notes without it are steady, independent of length or content.

The cause is a re-injection cycle rather than anything about the widget's size or position.

TaskNotes injects the card as a **direct child of `.markdown-preview-sizer`** (`insertAfterMetadataOrHeader`, `TaskCardNoteDecorations.ts`). Obsidian's reading view is virtualised, and every `updateVirtualDisplay` ends with:

```js
sizerEl.setChildrenInPlace([pusherEl, ...shownSections])
```

`setChildrenInPlace` removes every child not in that list, so **Obsidian deletes the card on every render pass**. Since 4.12.0, `src/editor/ReadingModeWidgetObserver.ts` re-injects it on the next animation frame. During a scroll that cycle runs continuously, and `updateVirtualDisplay` finishes by re-asserting `previewEl.scrollTop` each pass. That re-assertion is what is seen as the note jumping.

**Isolated by a single-variable test** on one note (125 sections), driving an identical scroll of 14 steps of 220px each time and counting `childList` mutations on the sizer:

| Condition | Jumps? | Sizer mutations | Scroll corrections |
|---|---|---|---|
| `showTaskCardInNote: false` | No | 0 | 0 |
| Card on, as shipped | Yes | 110 | 4, about 16px each |
| Card on, forced to zero height by CSS | **Yes** | - | - |

The third row is the one that isolates it. With `height: 0; overflow: hidden` and margins, padding and border zeroed, the card contributes exactly as much layout as it does when the setting is off, and it still jumps. So this is not a sizing or reflow problem - it is the DOM churn.

### Two things that look like the cause and are not

Offered because both are natural first guesses and cost me an afternoon.

**Section height accounting.** `measureSection` computes a section's height as `el.nextSibling.offsetTop - el.offsetTop`. The card sits between the header section and the next one, so its height is counted in neither, and the model runs short. This is real, but far too small to be the symptom: total model error peaks at about 26px roughly three seconds after a note opens and settles to 4px, while a non-task control note measures 0px across 1,100+ samples. The jumping continues while the model is within a few pixels of correct.

**Layout shift.** There are **no `layout-shift` performance entries at all** during the jump. The content is not moving; the scroll position is being re-asserted against the renderer's own model.

### The obvious fix does not work, and here is the measurement

Nesting the widget inside `.mod-header.mod-ui` looks like the answer, and half of it is:

- `setChildrenInPlace` only manages **direct** children of the sizer, so a nested widget is never deleted by it.
- The re-injection guard is `sizer.querySelector(widgetSelector)`, and `querySelector` is recursive, so a nested widget still satisfies it and no re-injection is scheduled.
- Prototyped live: the card renders correctly at 63px in its new home, and a single move places it.

But `.mod-header.mod-ui` is itself `sections[0]`, and **Obsidian detaches sections that scroll out of the render window**. Scrolling past the top of the note removes the header and the nested widget with it; the guard then finds no widget anywhere in the sizer and re-injects as a direct child, restarting the cycle. Measured on the same scroll: **87 sizer mutations, 6 widget add/remove events, and 2 scroll corrections of 79px** - worse per event, because the widget's height now perturbs the header section's measurement as well. The add and the remove land in the same microtask batch, so a `MutationObserver` cannot intervene between them.

So the widget needs a host that is neither a direct child of the sizer nor inside a virtualised section. Within the sizer only `pusherEl` survives every pass, since it is always in the keep-list, but it is a 1x0.1px box.

### Suggested directions

Offered as starting points, not prescriptions:

1. **Do not re-inject while a scroll is in flight.** The cycle only hurts because it runs per-frame during scrolling; coalescing re-injection until the scroll settles would remove the symptom without moving the widget.
2. **Distinguish "removed because offscreen" from "removed because the DOM was rebuilt."** The observer currently treats any absence as a reason to re-inject. Obsidian's own virtualisation is the commonest cause of absence, and re-injecting against it is what creates the loop.
3. **Render outside `.markdown-preview-sizer`.** Nothing outside it is virtualised, though the widget would then be pinned rather than scrolling with the note, and the guard would need changing to match.

### Relationship to #2081

This is the same underlying issue as [#2081](https://github.com/callumalpass/tasknotes/issues/2081) wearing a different face, and not a complaint about that fix. Before 4.12.0 the widget was deleted by `setChildrenInPlace` and simply stayed gone, which is #2081's reported symptom. `ReadingModeWidgetObserver` made it come back, which is correct as far as it goes, but the deletion still happens on every pass, so the widget now churns instead of vanishing. Fixing the deletion rather than the absence would close both.

[#1436](https://github.com/callumalpass/tasknotes/issues/1436) ("Error injecting task card widget in reading mode") and [#513](https://github.com/callumalpass/tasknotes/issues/513) ("Subtasks are not showing in Preview Mode") both look like the same root cause, from the pre-4.12.0 side.

### Steps to Reproduce

1. Confirm `showTaskCardInNote: true`.
2. Open a note tagged `#task` that is long enough to scroll, in **reading mode**.
3. Scroll down through it. The content jumps repeatedly.
4. Open a note of similar length **without** `#task`, in reading mode, and scroll. It is steady.
5. Turn **Show task card in note** off and repeat step 3. The jumping stops.
6. To rule out size as the variable, turn it back on and add a snippet: `.tasknotes-task-card-note-widget { height: 0; min-height: 0; margin-block: 0; padding: 0; border: 0; overflow: hidden; }`. The card now contributes no layout at all, and step 3 still jumps.

Live Preview is unaffected throughout, on every platform tested.

### TaskNotes Debug Info

```text
SYSTEM INFO:
	Obsidian version: 1.13.7
	Installer version: 1.13.4
	Operating system: Darwin Kernel Version 25.6.0: Sat Jul 11 15:26:29 PDT 2026; root:xnu-12377.161.13~4/RELEASE_ARM64_T8132 25.6.0
	Login status: logged in
	Language: en-GB
	Catalyst license: none
	Insider build toggle: off
	Live preview: on
	Base theme: adapt to system
	Community theme: Obsidian gruvbox 0.2.1
	Snippets enabled: 3
	Restricted mode: off
	Plugins installed: 20
	Plugins enabled: 16
		1: Image Converter v1.4.6
		2: Calendar v1.5.10
		3: Dataview v0.5.68
		4: Templater v2.25.0
		5: Natural Language Dates v0.6.4
		6: Metadata Menu v0.8.12
		7: Mermaid Tools v1.4.1
		8: Excalidraw v2.26.4
		9: Local REST API with MCP v5.1.0
		10: Note Toolbar v1.34.13
		11: Local Images Plus v0.16.4
		12: Another Quick Switcher v14.7.0
		13: Highlight active folder section v1.5.0
		14: SD Vault Tools v4.118.1
		15: JO Vault Tools v1.34.0
		16: TaskNotes v4.12.4

RECOMMENDATIONS:
	Custom theme and snippets: for cosmetic issues, please first try updating your theme and disabling your snippets. If still not fixed, please try to make the issue happen in the Sandbox Vault or disable community theme and snippets.
	Community plugins: for bugs, please first try updating all your plugins to latest. If still not fixed, please try to make the issue happen in the Sandbox Vault or disable community plugins.
```

### Affected Task Frontmatter

```yaml

```

```yaml

```

_No response_

### Console Errors or Logs

```text

```

_No response_

nelsonlove:
Fixes #2255.

## The problem

Reading mode is virtualised, and every render pass ends with `sizerEl.setChildrenInPlace([pusherEl, ...shownSections])`, which deletes any direct child of `.markdown-preview-sizer` that Obsidian did not put there. The task card is injected between sections, so it is deleted on every pass and re-injected on the next frame. During a scroll that cycle runs continuously and drags the reader down the note until it hits the bottom.

The scroll quiet period added in b29a29e6 defers re-injection, but it cannot stop this, because the problem is *where* the re-injection lands rather than how often it happens. Once the reader has scrolled past the top of the note, Obsidian has detached `.mod-header.mod-ui` too, so `getMetadataOrHeaderInsertionReference` falls through to the preview pusher - whose next sibling is the first section that is **currently rendered**, not the first section of the note. The card is therefore inserted into the middle of the text being read, and shoves the page down by its own height.

Measured on 4.12.5 by wrapping the scroller's `scrollTop` setter and capturing a stack on every write. One caller dominates:

```
e.updateVirtualDisplay  <-  e.onScroll
```

and the corrections are a constant step that repeats to the bottom of the note. The step equals the card's own occupied height, on each note:

| note | card occupied height | correction step |
|---|---|---|
| A | 579px | +578 / +579 |
| B | 690px | +690 / +698 |

The corrections arrived every 218-262ms, which is `DEFAULT_SCROLL_QUIET_PERIOD_MS` plus a frame. The quiet period did not stop the cycle; it set its tempo.

## The change

`setChildrenInPlace` only manages **direct** children, so nesting the card one level deeper leaves it untouched by the render pass. It is now placed inside the header, after the properties block.

Issue #2255 raised this route and found it insufficient on its own, which was correct: `.mod-header.mod-ui` is itself `sections[0]`, so Obsidian detaches it - with the card inside - when the reader scrolls past the top, and the observer then re-injects as a direct child and restarts the cycle. Two supporting changes close that:

- **The observer skips re-injection while the header is not rendered.** That detach is Obsidian's own virtualisation and must be left alone rather than fought. The card is restored when the header comes back, which is exactly when it can be seen.
- **Injection anchors on the properties block, not the end of the header.** Obsidian builds the header in stages, so appending during an early pass put the card above the properties, which then rendered underneath and shoved it down - a visible jolt on every note open.

Nesting also corrects the height accounting noted in the issue: `measureSection` takes a section's height as the gap between its own `offsetTop` and its next sibling's, so a widget between sections is counted in neither. Inside the header it is counted in that section.

## Tests

`tests/unit/editor/MarkdownWidgetInsertion.test.ts` pins the mechanism rather than the implementation, by simulating a render pass:

- a widget placed between sections is deleted by it
- a widget nested in the header survives it
- with the header detached, the old insertion reference resolves to the first rendered section - the trap this fixes
- with the header present but the properties block not yet rendered, injection is refused rather than landing above it

33 tests pass across all 6 editor suites.

## Verification

Built and run against a real vault on 4.12.5. The runaway is gone, the card holds its position under scrolling, and it renders once in its final place with no jolt.

## Trade-offs, stated plainly

- **The card is absent while the reader is scrolled past the top of the note.** It is inside the header, and Obsidian detaches that section. This is invisible in practice - the card lives at the top, so it cannot be seen from further down - but it is a real behaviour change from "always in the DOM".
- **Requiring the properties block** means a task note that renders in reading mode with no `.metadata-container` at all would not receive a card. If that case exists, a fallback to the end of the header once the header is known to be complete would cover it.
- **The `:has()` selector** that trims the properties block's bottom margin needs Chromium 105+. Where it does not apply, the gap is simply larger; nothing breaks.

Spacing is scoped to `.markdown-preview-view .mod-header`, so Live Preview and canvas are unchanged. The gap below the card uses `--p-spacing`, the note's own gap between blocks.

Happy to reshape any of this - the placement question in particular is yours to call.


callumalpass:
Thanks for the fix and detailed explanation. Merged into main—much appreciated!

